PR #5, opened 2014-09-30
QEP #3: Virtual Layers

mhugo:
This QEP covers the addition of "virtual layers" for QGIS, through the use of Spatialite


NathanW2:
Is there a status up on this.  After doing a lot of SQL this week I think having this in QGIS would be a great thing for the project.  Being able to query any layer, join, filter, etc would be a big win

mhugo:
So, for an update : we have been founded for a first version here : https://github.com/mhugo/qgis_vlayers 
I am still waiting for my customer to confirm everything is ok before making it "official". The github page gives some details.
Compiled binaries for windows (32bits only) can also be found here : https://github.com/mhugo/qgis_vlayers_bin/blob/master/qgis_28_vlayers_2_win32.zip?raw=true

I will be happy to have feedback from people here.

mhugo:
For information, I will be back on this topic soon. We have some users of the C++ plugin that would like to see it integrated into the core.
I will probably start by cleaning this document up considering what have been done during the implementation of the C++ plugin

NathanW2:
QEPs done via issues now.

New ticket at #44 

Sorry for the inconvenience.

Email out soon on why.

PR #7, opened 2014-10-05
QEP #5: stricter unit tests

nyalldawson:
This proposal is aimed at providing more stable releases and preventing regressions in QGIS releases by expanding the coverage of unit tests and adopting a hard line approach to compulsory unit tests.

PR #9, opened 2014-10-17
QEP #7: Rename composer

nyalldawson:
This proposal is regarding a change in terminology from "composition" -> "print layout".


NathanW2:
QEPs done via issues now.

New ticket at #43 

Sorry for the inconvenience.

Email out soon on why.

PR #8, opened 2014-10-10
QEP #6: Advanced digitizing tools

3nids:
Advanced digitizing tools in QGIS application.

See pull request https://github.com/qgis/QGIS/pull/1624

PR #10, opened 2014-11-03
QEP #8: Geometry redesign

mhugent:


wonder-sk:
May I suggest to use iterator pattern for various methods that currently return (possily nested) containers? I mean methods like
QgsAbstractGeometryV2::coordinateSequence()

mhugent:
@wonder-sk: to iterate over the vertices of a geometry, there is the method QgsAbstractGeometryV2::nextVertex( QgsVertexId& id, QgsPointV2& vertex ), which acts like a cursor.

mhugent:
Can we come somehow to a conclusion (e.g. decide the class architecture is ok)? I've continued developing the code in the branch and now that feature freeze is over, it will be good to think about how and when to integrate into the master branch.
I'm aware that not everything is implemented yet (e.g. geometry clipping and validation in edit tools is not yet there). And that the GUI most likely needs changes (e.g. edit tools). An integration in several steps will be good, e.g. in a first step only replace current QgsGeometry class, in a second step  change rendering code, in a third step edit tools. And in version 3, try to get rid of the compatibility class.

mhugent:
@NathanW2 : can we vote on this pull request soon?

NathanW2:
We can.  What is the status on the comments on the design?

mhugent:
>We can. What is the status on the comments on the design?

The only thing were I changed my mind was the usage of the visitor pattern. Because the v. pattern makes it difficult to add new geometry classes (and it would create an API break for each), I now think the existing virtual functions in QgsAbstractGeometryV2 should not be replaced with visitors.

pcav:
+1
thanks

wonder-sk:
Hi Marco

would it be possible to add few more details to the QEP before going with voting:
- how are the geometry data going to be stored internally?
- which classes in the hierarchy are abstract?
- what about QgsMultiLineString / QgsMultiPolygon classes - I can't see them in the hierarchy
- why are QList containers used instead of QVector containers?
- would it be possible to have a code examples in Python demonstrating the new geometry API? e.g. code that reads features and dumps geometry contents
- what are the plans for migration in QGIS 3.0 (i.e. what will happen with existing classes and proposed classes?)
- what about implicit sharing of geometries? Especially in the Python code, it is currently a pain that QgsGeometry class is not implicitly shared (they need to be explicitly copied otherwise they will refer to invalid memory). Can we make the new classes implicitly shared?

wonder-sk:
> I'm aware that not everything is implemented yet (e.g. geometry clipping and validation in edit tools is not yet there). And that the GUI most likely needs changes (e.g. edit tools). An integration in several steps will be good, e.g. in a first step only replace current QgsGeometry class, in a second step change rendering code, in a third step edit tools. And in version 3, try to get rid of the compatibility class.

No worries about not having the implementation complete - after all the QEP is a discussion about the design, not implementation...

wonder-sk:
> The only thing were I changed my mind was the usage of the visitor pattern. Because the v. pattern makes it difficult to add new geometry classes (and it would create an API break for each), I now think the existing virtual functions in QgsAbstractGeometryV2 should not be replaced with visitors.

Why is it difficult to add new geometry classes and why would it create API break? IMHO the virtual methods in the visitor class for a newly added class may have default (empty) implementation, thus not breaking anything.

I wouldn't be worried about addition of new geometry classes too much anyway - or do you envision that more geometry classes will be commonly added?

mhugent:
Many questions... I'm starting with the straightforward ones:
>how are the geometry data going to be stored internally?
One of the advantages now is that for each class, the storage can be decided individually. E.g. QgsLineStringV2 uses QPolygon for x/y and QVector for z,m
>which classes in the hierarchy are abstract?
QgsAbstractGeometryV2, QgsSurfaceV2, QgsCurveV2
>what about QgsMultiLineString / QgsMultiPolygon classes - I can't see them in the hierarchy
You can see them in the source (QgsMultiPolygonV2, QgsMultiLineStringV2). They have been introduced relatively late, because most of the functionality is implemented in the geometry collection class
>why are QList containers used instead of QVector containers?
QgsLineStringV2 and QgsCircularStringV2 are using QVectors to store the vertices. QLists are used e.g. in QgsCompoundCurves for the curve parts or in the multi-types for the pointers to the parts. But the the numbers of parts are usually not huge numbers...
>what about implicit sharing of geometries? Especially in the Python code, it is currently a pain that >QgsGeometry class is not implicitly shared (they need to be explicitly copied otherwise they will refer >to invalid memory). Can we make the new classes implicitly shared?
I didn't know this is a problem in Python, therefore did not plan to add implicitely sharing so far. But I still don't understand what the problem is. After all, the new classes are just normal classes. If the problem was the wkb pointer, that is now gone with the V2 classes.

wonder-sk:
> I didn't know this is a problem in Python, therefore did not plan to add implicitely sharing so far. But I still don't understand what the problem is. After all, the new classes are just normal classes. If the problem was the wkb pointer, that is now gone with the V2 classes.

For example if you would like to make a list of geometries of a layer - you would do something like this - please note the extra QgsGeometry():
```
lst = []
for f in layer.getFeatures():
  lst.append( QgsGeometry( f.geometry() ) )
```
The problem is that unless you make explicit copy of the geometry, the list ends up with references to geometries which are invalid (they get deleted in the loop).
But it's not just about python - also in c++ there are various cases when geometries are copied and implicit sharing would avoid the need for deep copying (in some providers or caching).

I realize that this is not something you may have considered in the design or implementation, and it is something that can be done only with API break in QGIS 3.

wonder-sk:
Btw. regarding the geometry access from QgsFeature - are the geometries still going to be accessed via f.geometry() or will there be a new method to get access directly to QgsAbstractGeometryV2 ?

mhugent:
>Btw. regarding the geometry access from QgsFeature - are the geometries still going to be >accessed via f.geometry() or will there be a new method to get access directly to >QgsAbstractGeometryV2 ?

At the moment, you still get the QgsGeometry (compatibility) class from QgsFeature. From QgsGeometry, you have access to the QgsAbstractGeometryV2. But it is a good idea to add a  method which gives back the new one directly. Like that, client code can already be prepared for the new interface.

mhugent:
>I realize that this is not something you may have considered in the design or implementation, and >it is something that can be done only with API break in QGIS 3.

The compatibility class QgsGeometry uses copy on write, so is your example going to work with it? And for the new classes, we can implement cow as we see fit (even without breaking the current v2 I guess).

mhugent:
>There are two aspects involved in the visitor pattern discussion: 1. methods virtual in >QgsAbstractGeometryV2 or outside, 2. if outside, visitor pattern or not

To the question of visitor pattern or not: it seems the v. pattern is very usefull if a object structure can be nested in a very complex way (example of parse tree in gof book). In that case, it is very handy for a visitor that it just has to implement visitxxx for every type in the structure and those methods are called from the object structure in the proper order. But in the case of mmsql geometries, the nesting is very simple. So wether you implement a visitor object (implements a visit-method for each geometry type) or a non-visitor object (has type switch and handler method for each geometry type) is about the same complexity. It can be argued that the visitor has the nice property to ensure a subclass must implement handling for all geometry types. But then again, with default methods in the base class, this does not hold true.

Question if virtual method or not: I do have plans to add support for Bezier splines at some point, as e.g. cartographers work frequently with this kind of geometry representation. And there are other possibilities of geometry types, e.g. clothoids for traffic planning. That's why I like to have certain methods as virtual in QgsAbstractGeometryV2, to ensure they must be implemented when adding a new type. Still, with geometryV2, it  will be possible to write classes implementing geometry behaviour by type switch. This can be a fallback for experimental code or plugins (but is not an ideal situation).

mhugent:
It is now possible to directly get an instance of the new Geometry class by calling  QgsFeature::geometryV2(). So to dump the geometry content (in python), use code like this:

print feature.geometryV2().asWkt()

To create geometries in Python, do e.g. (for a linestring)

lineCoords = [QgsPointV2(100,100), QgsPointV2( 110, 100 ), QgsPointV2( 110, 110 )]
lineString = QgsLineStringV2()
lineString.setPoints( lineCoords )

A zm-Geometry can be created by having zm points in the coordinate list:

lineCoords = []
lineCoords.append( QgsPointV2( QgsWKBTypes.LineStringZM, 100, 100, 23, 1.0 ) )
lineCoords.append( QgsPointV2( QgsWKBTypes.LineStringZM, 110, 100, 33, 2.0 ) )
lineCoords.append( QgsPointV2( QgsWKBTypes.LineStringZM, 110, 110, 43, 3.0 ) )
lineString = QgsLineStringV2()
lineString.setPoints(lineCoords)

mhugent:
I'd like to bring at least the geometry core into 2.10 (no editing tools yet). Please, can we have a vote about that QEP?

NathanW2:
Hey Marco,

That looks pretty good to me.  Happy that we can avoid API breaks, even if
it does mean adding V2 to the end.  We can always make a wrapper anyway if
need be.

Leave it with me I will do a vote call tonight if there is no other
objections. Sound good?

- Nathan

NathanW2:
Marco can I get you to update the the QEP document with the questions and
answer from the comments section.   Best to assume that no one will read
the comments after the merge.  I would like to setup something to also
publish QEPs to the website so everything needs to be in a single place for
reference.

mhugent:
@NathanW2 : do you mean to add a question / answer section to the QEP document? If yes, I can do. However some points in the discussion could be hard to get into the Q/A-scheme.

NathanW2:
Yes as much as you can.  Things like the Python example should also be included.  If someone reads it they need to be able to see the rationale without reading the comments.

mhugent:
@NathanW2: done

mhugent:
@NathanW2: vote call?

NathanW2:
Yep. Will do now.

NathanW2:
QEPs done via issues now.

New ticket at #40 

Sorry for the inconvenience.

Email out soon on why.

PR #11, opened 2014-11-16
QEP #9 - Composer rewrite (layouts)

nyalldawson:
This QEP covers a rebuild of the current composer functionality, to overcome design limitations with the current composer code

PR #13, opened 2014-11-27
QEP #11 - Fields and forms redesign

m-kuhn:
The tab *Fields* on the vector layer properties dialog should be split into two different tabs that separate data-related from form- and widget-related user interface elements.

This is expected to improve user experience and usability.


QEP-Fields-and-forms-redesign.rst
```diff
@@ -0,0 +1,133 @@
...
+- Form layout
+- Python init script for the form
+- ...
+
+Mockups
+~~~~~~~~~~~~~
+
+.. figure:: form1.png
+
+   Mockup of the *Form* tab with autogenerated layout. This will be shown embedded in the vector layer properties dialog.
```
dd-a: Hello,
I would like to add here a point I raised recently in the Qgis-user mailing list, i.e. the possibility to hide the url/path (same with 'web view') of a photo bellow it in the form.  It should be possible to have a tab (made with the Drag and drop designer) to simply display a picture, and eventually another field somewhere to navigate to the file.  Or to have one and not the other, which means have the possibility to choose to hide/show independently the picture and its address.  Currently (QGis 2.8.2) even when the field is set as 'non editable' the url is still shown, which I find quite useless most of the time.

dd-a: Another point appears more important to me: the possibility to set default values in the form.  From the discussion on Qgis-user it appears that we should have the possibility to set a default value and decide if it precedes or not the previous value entered.  IMO, by default the default values precede the last inserted values (otherwise defining default values would be somehow useless), except where specified otherwise in the field definitions.  So there would be an option in each field definition (a checkbox, next to the default value definition field) stating that the last non null inserted values do precede the default one.  But others consider that the last value should have precedence.  In any case this should be set at the level of the field.  I don't know if it should also be possible to specify in this order of precedence the default values provided by the database (i.e. when using a layer stored in PostGIS).

m-kuhn: Default vs "last used" vs provider defaults is an important point.
What do you think about this:

 1. By default use the provider defaults
 2. Allow overriding the provider defaults in the properties
 3. Allow "reuse last entered value" on a per field basis. The "last entered values" are currently session scoped, so the first time the default will be used, then the last entered value. (Or they would need to be saved in the project to be persistent).

I am thinking of the same for the "comment". By default use the one provided by the provider but allow to override it manually (Use-Case: allow localizing the comment based on the users locale).

Hiding the file path for the photo widget is unrelated. Please file this as a feature request in the issue queue, independent of this QEP.

What could be part of this QEP on the other hand is to allow multiple widgets for the same field with a different configuration.

dd-a: @m-kuhn I filed these two topics (default values, and hiding the photo path) already with links to this page.  Feel free to remove my photo path proposition from this page, or I can do it myself, or leave it.  Im' sorry for the mismatch.
Your description of the Default/last entered values looks good to me.
I have one more suggestion: the possibility to make a field mandatory, which if I'm not mistaken is currently different from not allowing null values.  At least, there should be a message saying 'This field is mandatory'.  This adds another choice to the Default value list again, the possibility to set NULL by default. 

I also had thought about having multiple widgets for the same field, but didn't dare to propose it because I expected it to be too complicated, either to implement or to use.  On the other hand this might lead to a reorganization of the widgets to avoid some apparent redundancy in the existing ones (like 'File Name', 'Photo' and 'Web View').

m-kuhn: Do you think there should be the possibility to add NULL as default to NOT NULL columns in order to force a user to make a choice?

dd-a: Sorry, forget about this, there's no need to specify NULL as a default value, it's already the result when nothing is chosen.  It's stupid.  I was confused because with dropdown lists, even when 'Allow null value' is not checked, it is still possible to chose nothing (by simply not touching the dropdown) and end up with a NULL ! This is not coherent and I will file a ticket for it. 

But it would be very useful the set a field as mandatory, meaning that something has to be entered, and therefore there should be a control when saving the form.

m-kuhn: For me mandatory and "NOT NULL" are the same. Do you agree?

At the moment, there is no coherence because if NOT NULL is an option depends on the widget type while it belongs on the field level and widgets should respect the configuration of the underlying field (this change is in the scope of this QEP).

dd-a: Yes I agree :-)

ninsbl: Regarding the "default values": It would be nice if they could be calculated, e.g. based on other attributes which would be a kind of "trigger" like option.
Here it would be splendid if one could use variables (e.g. %USER% or expressions like in the field calculator or Relation Reference widget.

 "Calculated" might probably be a new widget type... ?

Also cascaded logic as an extension to current value relations would be nice, like demonstrated here: http://medspx.fr/blog/Qgis/cascade_forms/

Should this go into a feature request in the issue queue?

BTW, do QGIS forms already support image previews for images stored in e.g. PostgreSQL, as well as image upload to DB through a form (not only the path)?

m-kuhn: Calculated default values would be a good enhancement. I think that can be put into this package or added in a separate package.

Calculated as widget type: wouldn't that be what virtual fields already do?

Cascaded logic : introduced in 2.10 -> https://www.qgis.org/en/site/forusers/visualchangelog210/index.html#feature-filter-chaining-for-relation-reference-widget

No, images (and BLOBS in general) are not yet supported by QGIS.

ninsbl: Thanks for the info.
My idea of "Calculated" as widget was indeed nonsense. A line edit widget with the possibility to calculate default values would do what I had in mind...

NathanW2:
QEPs done via issues now.

New ticket at #37 

Sorry for the inconvenience.

Email out soon on why.

PR #16, opened 2015-01-05
QEP #13: Efficient Snapping and Geometry Queries

wonder-sk:


qep-13-efficient-snapping-and-geometry-queries.rst
```diff
@@ -0,0 +1,234 @@
...
+------------------------
+
+It is expected that snapping performance will be sped up significantly.
+From some quick tests, the current snapping took ~30ms to find the closest vertex, while with new implementation
+needed only less than 1ms. This is because current snapping needs to hit data provider (if the layer is not in editing mode).
+
+There is some cost in initial indexing in ``QgsPointLocator``. For a layer with ~50K points this took about 100ms.
+This is just one-time cost to build the index from data provider's features when queries are first needed.
+No extra cost when map is rendered.
+
```
m-kuhn: Will the whole layer be cached?
Is there a mechanism that avoids out-of-memory errors?

qep-13-efficient-snapping-and-geometry-queries.rst
```diff
@@ -0,0 +1,234 @@
...
+   
+       1. snap to current layer (with default snapping type and tolerance)
+       2. snap to all layers from configuration (advanced)
+   - not directly connected to map canvas - need map settings and current layer to be explicitly set
+   - optional snapping on intersections
+   - client code can access locators
+   - supports filtering of matches (e.g. accept matches only from a particular feature ID) with ``MatchFilter``
+   - configuration can be read from project (but does not need to be)
+   - replacement for ``QgsMapCanvasSnapper``
+
```
m-kuhn: For a plugin I was looking for a possibility to do interactive snapping with priority

 * If there's a point on layer A, snap to that
 * If there is none, try to snap to layer B

I always would just want a single result (closest).

Do you think this is possible? Can ``QgsSnappingUtils`` be subclassed to introduce that logic?

rduivenvoorde:
Hi Martin,

do I read that you keep the geometry indexes in memory?

please also consider the data-sources (like big postgis db's) which you just cannot keep into memory.

Currently we cannot open attribute-table anymore on this kind of data sources: http://hub.qgis.org/issues/10619 it was downgraded as a blocker, but to me it is :-(
We do not want to go into the direction where QGIS is only usable for smaller (file based) datasets?

mhugent:
Can the new snapping class be easily adapted to non-curved geometries for segment snap in future?

qep-13-efficient-snapping-and-geometry-queries.rst
wonder-sk: Yes it should be easy to implement such strategy. It could be a new "ordered" mode where the search would be stopped once first good match is found. I thought about using this strategy by default for QgsSnappingUtils, but the project's snapping configuration is not an ordered list so...

qep-13-efficient-snapping-and-geometry-queries.rst
wonder-sk: In the current implementation yes the whole layer is cached in memory (not automatically - only when first needed) - though it should be fairly easy to use spatial index library's disk storage instead of memory storage for big layers.

wonder-sk:
> do I read that you keep the geometry indexes in memory?
> 
> please also consider the data-sources (like big postgis db's) which you just cannot keep into memory.

As noted in the comment to Matthias, it should not be difficult to keep the cache on disk to keep the memory consumption low. I am downloading the dataset you linked to test how it will behave.

wonder-sk:
> Can the new snapping class be easily adapted to non-curved geometries for segment snap in future?

I am not sure I understand the question. It supports snapping to segments of non-curved geometries. Did you mean support curved geometries? In that case, yes, the adaptation should be easy, the tree leaves for segments will just need slightly different encoding and addition of point-to-curve distance test.

mhugent:
Ups, yes I meant curved geometries.
That means if someone adds a new geometry type, he needs to add switches, encoding, decoding, segment snap test for type in different places in QgsPointLocator, otherwise the geometry will not be snappable?

wonder-sk:
> That means if someone adds a new geometry type, he needs to add switches, encoding, decoding, segment snap test for type in different places in QgsPointLocator, otherwise the geometry will not be snappable?

Exactly. Adding support for new geometry types is actually easier that it may seem :-)

mhugent:
>Exactly. Adding support for new geometry types is actually easier that it may seem :-)

But it is also extremely easy to forget to implement that. In an object oriented world, it would be cool if the geometry could give back a list of abstract segments. The segment would need to implement encoding / decoding / test methods. And if the programmer forgets to provide that, the compiler will remind him.

wonder-sk:
> But it is also extremely easy to forget to implement that. In an object oriented world, it would be cool if the geometry could give back a list of abstract segments.

Generally I do not have a problem with such approach - but that looks like something that may be added after the major geometry refactoring. After all, it is a rare event when new geometry types are added and there are dozens of other places in QGIS code that need to be adjusted as well.

> The segment would need to implement encoding / decoding / test methods. And if the programmer forgets to provide that, the compiler will remind him.

Agreed with the "distance to point" method. But encoding and decoding for the purposes of storage in point locator R-tree is just an implementation detail for one small component - such code does not belong to a generic "abstract segment" class.

mhugent:
>After all, it is a rare event when new geometry types are added and there are dozens of other places >in QGIS code that need to be adjusted as well

Yes, but it does not hurt keeping the number of places low (especially for newly added code).

>Agreed with the "distance to point" method. But encoding and decoding for the purposes of storage >in point locator R-tree is just an implementation detail for one small component - such code does not >belong to a generic "abstract segment" class.

True, this functionality is not generic to a segment. But maybe it is something that a QGIS-Segment needs to provide to be of use. What place do you suggest to put the code in?

wonder-sk:
> But maybe it is something that a QGIS-Segment needs to provide to be of use. What place do you suggest to put the code in?

Currently I can't imagine where else such code would be useful - so the QgsPointLocator is the only place I can think of.

PR #20, opened 2015-03-26
add enhanced metadata management and MetaSearch publishing QEPs

tomkralidis:


haubourg:
Hi , I discover your QEP just now.
At my job, we currently begin to harvest qgs and qlr files to generate metadata ISO 19139, DC and ISO 19110 xmls, and push them to a Geonetwork catalog, using CSW-T too. We use Talend metadata crawler job to do that, which allows data monitoring to keep metadata up to date. 
Working with qgis server in the same time, we got conscious that there is confusion in QGIS between ISO 19xxx norms and WxS metada services.  For instance, Title metadata for layers must be unaccented and well formed to be used in WxS Url's, when a metadata user expects there a quite long and explicit title for metadata record. So the very same field can break WFS requests if user types uncontrolled text in it?
The same goes for project properties.  So, we should start a cleanup work in QGIS to clarify the norms used behind each metadata fields, so that metadata generating use case and OGC WxS services do not collide. I expect to work on that by the end of the year and propo
Opinions on that? 
Cheers
Régis

NathanW2:
QEPs done via issues now.

New ticket at #33 

Sorry for the inconvenience.

Email out soon on why.

PR #21, opened 2015-06-19
Add QGIS server access control interface for python plugin

sbrunner:


elpaso:
A couple of random comments/questions:
I suggest to rename your "plugins" in "filters", this is the same naming convention we have with the existing plugins system: a server plugin can register one or more QgsServerFilters, you should probably think about renaming existing QgsServerFilter into QgsServerRequestResponseFilter and rename your new  QgsAccessControlPlugin filters something like QgsServerAccessControlFilter. Calling your filters "plugins" confused me a lot. Also, "security" in server interface should probably be "accessControl".

Also: wouldn't be best to subclass QgsServerFilter? Cant you just add more hooks for AC to the QgsServerFilter class?

What are the problems if you add more methods directly into https://github.com/qgis/QGIS/blob/master/src/server/qgsserverfilter.h ? Can't you just add security filters there? Then they would be called where they are needed, using the very same system we are using for existing "requestReady" and similar.

something like:

```python
class HelloSecurityFilter(QgsServerFilter):

    def __init__(self, serverIface):
        super(HelloSecurityFilter, self).__init__(serverIface)

    def layerPublishPermission( self , layer):
        ... do whatever you want with this layer

    def requestReady( self):
        pass # do nothing

```

One of the main concerns I have that we are adding plugin's server specific code into core classes (like src/core/qgsmaprenderer.h). You should try to avoid this but I don't know if it is really possible.

Can you please provide a minimal example python code of a new AC filter and plugin?

sbrunner:
> Also: wouldn't be best to subclass QgsServerFilter? Cant you just add more hooks for AC to the QgsServerFilter class?
> ...

Too reason why I do this:
 * I like to separate the different kind of server plugin 
 * little performance issue: If we register a plugin with only the method `requestReady` and `sendResponse` we enter in all the code for the accessControl and it can create a slow down, but it's true that's probably negligible...

Finally if you think that's better it's not an issue fro me to do it :-)

luca76:
It would be useful if we can limit the access on:
 - QGIS projects
 - layers
 - specific layer attributes

sbrunner:
@luca76 
For the layers and features(layer) attributes it's already expected
For the "QGIS projects" you should be able to access to the query string than you can catch the map argument :-)

elpaso:
@sbrunner 
I think that the performance issue would be really negligible while we would have less code to maintain and to document and a coherent API, so my advice is definitely to add more methods to QgsServerFilter instead of having a separate class.

Any idea about how to overcome what I feel is the main issue (server-specific code into "core" library)?

sbrunner:
OK, thanks @elpaso for your response :-)

elpaso:
@luca76 
as @sbrunner said, limiting access to projects and layers as a whole is already possible **now** with the existing plugin filters implementation. The proposed enhancement is about adding row-level (single feature and attributes) authorization through spatial operators (BBOX as a start).

NathanW2:
QEPs done via issues now.

New ticket at #32 

Sorry for the inconvenience.

Email out soon on why.

PR #23, opened 2015-10-20
Add QEP for Unit testing and "CRITICAL" classes

nyalldawson:
Proposal for the introduction of compulsory unit testing for "CRITICAL" classes, and details the requirements for a class to be categorised as CRITICAL.


nyalldawson:
This supersedes #7

timlinux:
Personally I preferred QEP #5. The approach outlined here requires much more insight into the runnings of the project and the state of the test suite. Having a blanket rule rather that says 'no tests, no merge' is much simpler to enforce and for contributors to understand.

In InaSAFE we use an 80% coverage requirement for tests based on the premise of diminishing returns - to test the last 20% of coverage requires a much higher effort as typically those will be edge cases like platform / data source / locale etc. specific behaviours which require a lot of work to make a replicable test. I think it would be good to follow suite in QGIS. And we have a simple assertion that no contributed code should lower the overall coverage of the test suite.

NathanW2:
I think that makes sense when you have a testable code base.  A lot of QGIS
is pretty hard to test and we just need to build it up.

timlinux:
> I think that makes sense when you have a testable code base.  A lot of QGIS
is pretty hard to test and we just need to build it up.

Are you referring to the coverage requirement here or the mandating of tests. Agreed on coverage, we can't expect this from work go. WRT to mandating tests, I can't see any reason why PR reviewers can not simply label a PR as 'needs tests' and hold the merge until that is taken care of...

mhugo:
This CRITICAL classification is interesting to prioritise addition of unit tests to existing classes. But I also prefer the simple rule "no test, no merge".
Codes that are hard to test : GUI and probably providers (but not even sure). Do you see other parts ?

> Requiring all changes to classes which satisfy 2.1.1 to have unit tests would likely push away potential contributors who do not have the time or skill required to write unit tests from scratch

I am not sure to agree with that point. If a potential contributor has time and skill to write a new feature, is it really hard to write a unit test with it ? I don't think so. I am really not sure it would exclude contributions.

m-kuhn:
I like this approach. It is based on a well-defined *minimal required* test suite. Nobody will prevent anybody from writing further tests or requiring other pr's to be accompanied by a test if suitable (at least I will still attach the *requires tests* label where I see fit even if it is not enforced by this QEP)

Stricter approaches discussed here have some properties that make it hard to enforce them like it's hard to tell which part of a PR requires a test, people may write silly tests, just to fulfill the requirement and push the coverage and last but not least it seems to be politically difficult. QEP #5 has been around for a year and obviously did not make it so far.

In order not to get stuck in a "but we should have something stricter" situation and dismiss this PR I would like to strongly encourage everybody to see this as a real chance to increase test requirements. This QEP will scale with an increasing test suite and if we see fit for additional preconditions to mandatory tests in the future we will be perfectly able to put them into place at a later stage.

strk:
+1 to simple rule. "CRITICAL" is too debatable

nyalldawson:
Look... anyone who knows me will know that I'm a huge proponent of stricter unit testing in QGIS. But I honestly think this is a better approach than QEP #7  for where we are NOW with the project. We've got to start small and slowly work out way there, and this approach is a good start.

Note that I don't think that this and #7 are mutually exclusive. This QEP covers a proposal for really deep unit test coverage for certain areas of the codebase. For example, I think the coverage of the unit tests included in https://github.com/qgis/QGIS/pull/2189 are sufficient for general code to be included in core. But it's not sufficient for the critical classes described in this proposal. For those, I don't want to settle for anything less than 100% coverage (I'm talking about coverage by lines of code, not methods BTW... probably should clarify that in the QEP). The critical classes are for things we HAVE to get right... if we don't, we may as well pack up, close off the repos and buy an ESRI license ;) 


Some answers to other questions:

>I am not sure to agree with that point. If a potential contributor has time and skill to write a new feature, is it really hard to write a unit test with it ? I don't think so. I am really not sure it would exclude contributions.

I've personally seen at least two instances of a new contributor being asked to write tests for a class which doesn't have existing tests, and seen the intense frustration that caused them as they struggled with trying to do this when there's no precedent (via IRC and a PR that I can't find right now).  As Matthias mentioned, we can still block PRs which don't have sufficient testing for other classes.

>The approach outlined here requires much more insight into the runnings of the project and the state of the test suite.

Not really. Like it says in the QEP, when a class gets marked as CRITICAL, I'll cover the source with the described comments which explain that changes require unit tests. There's no extra need for deeper knowledge of the project ... the comment will be right there in front of the developer!

That said, I think that the QEP probably needs to explain how a class becomes critical. I'm thinking that when it meets the criteria one of the magic-flying-QGIS-pixie-monkeys (probably Matthias or I) will submit a PR with the CRITICAL tagging throughout the class for discussion.

....

There's another way I could have phrased this QEP... I'm going to do a bunch a work soaking certain classes in unit tests, and I want the PSC to grant me some authority to ensure that these classes STAY soaked in tests. QEP #5 and the current contributor guidelines are nice, but there's no power behind them. This QEP is about
1. extremely deep testing of certain classes
2. teeth to bite people who ignore this ;)

mhugo:
> Note that I don't think that this and #7 are mutually exclusive

Thanks for this clarification. So it does not supersede #7. Then, ok, adding a tag to state some code parts are critical and need deep test coverage seems good :)

luipir:
just a note, bug fixing that touch some methods not yet covered by tests (for a CRITICAL class)? The developer have to cover all the test not covered by previous developers? This is just to prevent bug fixer contribution due the risk of
"I've personally seen at least two instances of a new contributor being asked to write tests for a class which doesn't have existing tests, and seen the intense frustration that caused them as they struggled with trying to do this when there's no precedent (via IRC and a PR that I can't find right now). As Matthias mentioned, we can still block PRs which don't have sufficient testing for other classes."

nyalldawson:
>just a note, bug fixing that touch some methods not yet covered by tests (for a CRITICAL class)? The developer have to cover all the test not covered by previous developers?

@luipir No - that's why one of the requirements for a class to reach CRITICAL status is that it has to have 100% test coverage already. So a bug fix to a critical class will only require testing of the new/changed code, not all the existing code - those will already be covered by existing tests.

nyalldawson:
@NathanW2 I've finalised this proposal, can it be raised for voting now?

NathanW2:
@nyalldawson yes I think so.  I will send it out tonight.

NathanW2:
QEPs done via issues now.

New ticket at #42 

Sorry for the inconvenience.

Email out soon on why.

PR #24, opened 2015-10-21
Proposal for QGIS 3.0 release after 2.14

nyalldawson:
Proposal for an API breaking 3.0 release after 2.14 LTR release.

### DIFF
--- a/QEP-QGIS_3.rst
+++ b/QEP-QGIS_3.rst
@@ -1,0 +1,114 @@
+.. _qep#[.#]:
+
+============================
+QGIS Enhancement #: QGIS 3.0
+============================
+
+:Date: 2015/10/21
+:Author: Nyall Dawson
+:Contact: nyall dot dawson at gmail dot com
+:Last Edited: 2015/10/21
+:Status:  Draft
+:Version: QGIS 3.0
+
+1. Summary
+----------
+
+QGIS 2.16 should become QGIS 3.0, and include:
+
+- an API break
+- Qt5 only build support
+- Python 3.0 support
+- An extended development cycle to facilitate these changes
+
+2. Background
+-------------
+
+
+3. Proposed Changes
+-------------------
+
+3.1 API Break
+.............
+
+QGIS API has remained stable throughout the 2.0 series. While it is acknowledged
+that this is desirable for plugin and script maintainers, QGIS 3.0 will allow 
+API breaks. This is required to clean up the codebase and make deeper changes
+and bug fixes which are currently impossible to do without breaking API. It
+also allows for removal of old compatibility code (currently QGIS master has over
+300 methods marked as deprecated), simplifying future development and 
+maintenance efforts.
+
+3.2 Qt 5 Requirement
+....................
+
+Currently QGIS supports building on both Qt4 and Qt5 libraries, with the default
+build being Qt4. Qt4 has been end-of-lifed by it's developers and is no longer
+maintained. This raises signficant risks for QGIS if we do not move to Qt5 builds
+as quickly as possible. For instance, Qt4 cannot be built on the latest OSX 
+release (10.11). Third party patches have been made which partly resolve this
+issue, but the side effects are unknown. It is highly likely that a future
+OSX release, Windows update or linux packaging change will prevent use of
+Qt4 on one of our supported platforms and result in QGIS being unavailable
+for that platform.
+
+Additionally, numerous bugs present in the current QGIS release which are
+caused by issues in Qt have been resolved in Qt5. Moving to Qt5 is the only way
+to fix these issues.
+
+There's also significant performance and feature benefits made possible by
+the switch to the newer platform.
+
+3.2.1 Issues with moving to Qt5
+...............................
+
+Qt5 has dropped support for the QWebView/QWebPage classes. The replacement
+class, QtWebEngine, is immature and lacks several important features which
+are currently used by QGIS. The largest impact this will have will be on
+the composer "HTML" item. This item is currently rendered using a QWebPage
+painting directly on to a QPainter surface, resulting in text rendering
+as text objects and vectors. This is NOT currently possible with the 
+replacement classes, and it is unclear whether this will be brought back
+by a future Qt release. It is also impossible to render QtWebEngine contents
+using a transparent background. Unfortunately, there is no workaround for these
+limitations and the feature will have to be modified so that HTML content is either:
+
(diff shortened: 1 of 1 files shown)


QEP-QGIS_3.rst
```diff
@@ -0,0 +1,114 @@
...
+limitations and the feature will have to be modified so that HTML content is either:
+
+- rendered as an upscaled (read: pixelated) image
+- rendered using the subset of HTML and CSS supported by QTextDocument (see
+  http://doc.qt.io/qt-5/richtext-html-subset.html )
+
+3.3 Python 3.0
+..............
+
+** I'm unclear about the exact requirement here - but I believe PyQt5 is
```
NathanW2: Even if not I think it's best just to go Python 3 regardless as 2.7 is going end of life.

QEP-QGIS_3.rst
```diff
@@ -0,0 +1,114 @@
...
+..................
+
+Almost everything!
+
+#. Further Considerations/Improvements
+--------------------------------------
+
+Support for plugin developers in the form of guides for porting to QGIS 3.0
+will need to be developed.
+
```
NathanW2: There should also be some tools that we can use here to help.  I will see what I can find so we can make a list.

QEP-QGIS_3.rst
```diff
@@ -0,0 +1,114 @@
...
+4.0 Timeline for 3.0
+--------------------
+
+In order to maximise the value of an API break, and to allow sufficient
+time for the changes required to port to Qt5/Python 3.0, the 3.0 development
+cycle should span twice the normal development cycle. This means that
+the revised timeline for 2016 would be:
+
+2.14 (LTR) 26.02.2016
+2.16 - NO RELEASE
```
nirvn: I was thinking of an alternative proposal to avoid skipping a release cycle but still give us enough time to complete the port to Qt5.

Basically, we begin the development cycle of both 2.14 LTR _as well as_ 3.0 at the same time, following the public release of 2.12. The simultaneous development of those two versions would have two distinct objectives and release dates:
- for 2.14 LTR, the development cycle would last 4 months and would be released as planned on 26.02.2016 with a focus on stability and bugfix based on the feedback from users on 2.12 (small-ish new features could be pushed too)
- for 3.0, the development cycle would last 8 months, with the first 4 months overlapping with 2.14 LTR, and be released on 26.06.2016 (as our 4-month release cycle would have it) with a focus on porting QGIS to Qt5 and python 3 as well as significant new features.

This alternate timeline proposal both speeds up the delivery of QGIS 3.0 as well as insuring QGIS 2.14 LTR is shipped with a focus on stability as devs can push new features onto QGIS 3.0.

Thoughts?

QEP-QGIS_3.rst
```diff
@@ -0,0 +1,114 @@
...
+- rendered as an upscaled (read: pixelated) image
+- rendered using the subset of HTML and CSS supported by QTextDocument (see
+  http://doc.qt.io/qt-5/richtext-html-subset.html )
+
+3.3 Python 3.0
+..............
+
+** I'm unclear about the exact requirement here - but I believe PyQt5 is
+python 3.0 only, so our plugins would need to be ported to Python 3.0
+also....
```
m-kuhn: No, the reasons for me two bundle the two changes are: Qt5 as well as python3 require changes to plugins. Although updates of these libs could theoretically be split and performed independently, I think it's wise to bundle them, since plugin devs only have to upgrade their plugins once.

QEP-QGIS_3.rst
```diff
@@ -0,0 +1,114 @@
...
+Qt5 has dropped support for the QWebView/QWebPage classes. The replacement
+class, QtWebEngine, is immature and lacks several important features which
+are currently used by QGIS. The largest impact this will have will be on
+the composer "HTML" item. This item is currently rendered using a QWebPage
+painting directly on to a QPainter surface, resulting in text rendering
+as text objects and vectors. This is NOT currently possible with the 
+replacement classes, and it is unclear whether this will be brought back
+by a future Qt release. It is also impossible to render QtWebEngine contents
+using a transparent background. Unfortunately, there is no workaround for these
+limitations and the feature will have to be modified so that HTML content is either:
```
m-kuhn: QWebView is available for Qt5 on many platforms but not on all (e.g. mobile). I think it would be best to create an intermediate layer that allows to access both systems with unified methods. Which one is used is determined at build-time. At the moment we have QgsWebView that does not support the QtWebEngine backend, but the same thing with a second backend added would be great.

QEP-QGIS_3.rst
m-kuhn: What I wonder is if/how we should make it possible to maintain a plugin that is compatible with 2.x and 3 (at least with a reduced functionality). It's well possible to write python code that is compatible with Python 2 and 3, so that should be easy. Qt API's are also similar, we would probably only need a common "from PyQt import XXX" wrapper instead of "from PyQt[4|5] import XXX" to make it possible to have a large common foundation.

That would help people to prepare for the change while continuing to work on a single codebase that they can deploy to the current versions.

QEP-QGIS_3.rst
```diff
@@ -0,0 +1,114 @@
...
+and bug fixes which are currently impossible to do without breaking API. It
+also allows for removal of old compatibility code (currently QGIS master has over
+300 methods marked as deprecated), simplifying future development and 
+maintenance efforts.
+
+3.2 Qt 5 Requirement
+....................
+
+Currently QGIS supports building on both Qt4 and Qt5 libraries, with the default
+build being Qt4. Qt4 has been end-of-lifed by it's developers and is no longer
```
timlinux: end-of-life'd

QEP-QGIS_3.rst
timlinux: I like this suggestion @nirvn - having a long bug fix period for 2.14 would make the LTR release a lot more meaningful from the point of release. I would effectively freeze the features at 2.12 (with possibly small feature additions as you suggested).

QEP-QGIS_3.rst
NathanW2: Yeah I don't think there will be much work involved in making it work in both.  Just need to plan for it so that people can do it quickly and not feel like they have no idea what needs to change.

NathanW2:
QEPs done via issues now. 

**New ticket at #29**

Sorry for the inconvenience.

Email out soon on why.

PR #26, opened 2015-10-27
Advanced labeling

mhugo:
Here are some new QEPs related to advanced manual labeling.

It comes from user needs (QEP writing funded by Agence de l'Eau Adour Garonne).

This is a first version of the QEPs. I've tried to translate the user needs into generic enhancements to the core. I would like to collect feedback from reviewers regarding the proposed design and architecture, before focusing on implementation details, since there might be better / simpler ways to achieve the same goals.

The user needs are:
- to ease the way labels can be manually placed by the user. It requires currently to have additional fields to store X, Y (and angle) values for labels. That led me to propose a generic way to store "auxiliary" data (you're welcome if you have a better name) linked to a layer. See the [QEP document on auxiliary storage](https://github.com/mhugo/QGIS-Enhancement-Proposals/blob/adv_labeling/QEP-auxiliary-storage.rst)
- to allow the use of a simplified geometry for the curved labeling of waterways. This leads me to the [QEP on "label path"](https://github.com/mhugo/QGIS-Enhancement-Proposals/blob/adv_labeling/QEP-label-path.rst)
- to allow to precisely place isolated words or letters of a label. This leads me to the [QEP on "label annotations"](https://github.com/mhugo/QGIS-Enhancement-Proposals/blob/adv_labeling/QEP-label-annotations.rst)
- to allow to hide labels when a "mask" layer is used. The [QEP on "label mask"](https://github.com/mhugo/QGIS-Enhancement-Proposals/blob/adv_labeling/QEP-label-mask.rst) is about that.


dakcarto:
Hi Hugo,

Nice proposal for some needed features!

Can you put links, in your description above, to the individually rendered .rst files? So it is is easy to jump to the individual QEPs, like:

* [QEP-label-path](https://github.com/mhugo/QGIS-Enhancement-Proposals/blob/adv_labeling/QEP-label-path.rst) - to allow the use of a simplified geometry for the curved labeling of waterways.

QEP-auxiliary-storage.rst
```diff
@@ -0,0 +1,178 @@
...
+The proposed solution have the following requirements :
+
+- It should allow to transparently add columns of any type to a layer, without adding them to the source data source.
+Beside usual types (string, real, integer), we may also think of geometry types. We only consider types that can be converted to atomic types (string for WKT or WKB)
+The user should be able to edit values in these columns when the layer is in editing mode.
+
+- These values should not be stored with the source data, and should be carried by the project, i.e.
+when the project file is moved or copied, the auxiliary data should still be there.
+
+- Adding a new row to the layer also adds a new row for each auxiliary field (with a null or default value)
```
nyalldawson: what about only adding the row when data is present in one of the auxiliary fields, and have the default value returned for these fields as null? This would avoid the auxiliary table bloating out with effectively empty rows, and would keep the project file small and responsive even if aux fields are used with a table containing millions of features.

QEP-auxiliary-storage.rst
```diff
@@ -0,0 +1,178 @@
...
+
+- Removing a row from the layer also removes the corresponding row of each auxiliary field
+
+- New GUI elements should ease the creation of auxiliary fields
+  - on each "data-defined" property
+  - possibly on groups of "data-defined": for manual label placement or label connectors
```
nyalldawson: Maybe this could just be done as required too. Eg, if user goes to move a label using the move label tool, then the x/y aux fields are just instantly and linked to the data defined property without any user intervention.

QEP-auxiliary-storage.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Given the current state of the core code where lots of classes assume the project file is an XML file (everything that is saved to the project file
+has a writeXML method), it seems hard to completely refactor the project file format to a format based on SQLite.
+
+We propose to use a Zip file as the root of the new project file format. A Zip file exposes a filesystem-like structure.
+The original XML-based project file will then be embedded in the Zip file. And additional files could de added, and in particular
+an SQLite file.
+It only adds a step after actually saving the XML tree: including it in a Zip file. Symetrically during the loading, the first step would be to extract the XML file and pass it to the
+existing readXML methods.
+
```
nyalldawson: What's the expected timeframe for this? Changing the project file format sounds like a 3.0 feature to me.

QEP-auxiliary-storage.rst
```diff
@@ -0,0 +1,178 @@
...
+Core changes
+------------
+
+Project file
+------------
+
+The current project file format (XML) is too limited to carry an efficient storage of such auxiliary data. It should somehow be evolved
+to include a binary indexable format. SQLite is chosen as a first candidate for this data format. But it might evolve if performance issues
+are found.
```
nyalldawson: This could also be used for saving memory layers within a project.

QEP-auxiliary-storage.rst
```diff
@@ -0,0 +1,178 @@
...
+It only adds a step after actually saving the XML tree: including it in a Zip file. Symetrically during the loading, the first step would be to extract the XML file and pass it to the
+existing readXML methods.
+
+Auxiliary fields
+-------------------
+
+An auxiliary field is defined for a given layer. It is a new type of field: a new enum value for QgsFields::FieldOrigin (OriginAuxiliary). The code
+of QgsVectorLayer, QgsVectorLayerFeatureIterator and QgsVectorLayerEditBuffer will be changed to reflect this new type, so that they will appear in the list of fields of a layer (alongside native fields and virtual fields).
+
+Since an auxiliary field is conceptually a (left) join on an external database table, auxiliary fields may only be added to vector layers that have a **primary key**.
```
nyalldawson: How will this be conveyed to the user?

QEP-auxiliary-storage.rst
```diff
@@ -0,0 +1,178 @@
...
+Core changes
+------------
+
+Project file
+------------
+
+The current project file format (XML) is too limited to carry an efficient storage of such auxiliary data. It should somehow be evolved
+to include a binary indexable format. SQLite is chosen as a first candidate for this data format. But it might evolve if performance issues
+are found.
+
```
nyalldawson: Also, where will the sqlite file be stored while a project is being edited? I assume the user will be working with a temporary version of the sqlite database which is only stored inside the project zip on project save?

QEP-auxiliary-storage.rst
```diff
@@ -0,0 +1,178 @@
...
+For some particular use cases (computing geometries by expressions), such computations may be slow. Evaluating the expression that defines a virtual field is needed only when
+the fields it refers to change. We then propose to allow auxiliary fields to store values that are initially defined as expressions.
+
+An expression-defined auxiliary field can only be accessed for reading. The underlying fields refered to by the expression will be listed and when attributes of the vector layer
+is updated, linked expression-defined auxiliary fields will be updated.
+
+The value of an expression of complex type may need a serialization step before being cached. For instance, values of type QgsGeometry will be stored as WK(B|T).
+
+(Actually this cache mechanism for virtual fields is not specific to auxiliary fields and could be generalized to any tpyes of providers)
+
```
nyalldawson: This section is a bit unclear... are you proposing to alter the existing virtual fields implementation for this or are you proposing to add expressions to the aux fields? 

I can see a few issues with either approach:
- data may change outside of the project, eg if a layer is edited in a different QGIS project, or directly within a database. How will the field know it must be updated in this case?
- It's not always possible to know what fields a QgsExpression depends on (eg, using the "attribute" function). Or if a user-defined python based expression is used, the evaluated value could potentially depends on anything (other layers, external web services, ... )

I wonder if this section should be left out of this QEP and left for a future discussion. To me it seems unrelated to the proposed labeling work, and potentially may side track these discussions....

QEP-auxiliary-storage.rst
```diff
@@ -0,0 +1,178 @@
...
+QGIS Enhancement ??: Auxiliary storage
+========================================================================
+
+:Date: October 2015
+:Author: Hugo Mercier
+:Contact: hugo dot mercier at oslandia dot com
+:Last Edited: 
+:Status:  
+:Version:
+
```
nyalldawson: I like the general direction this proposal is going... it sounds a lot like how MapInfo embeds custom labeling properties within a workspace file, but without all the horrible performance issues MI's implementation suffers from.

QEP-label-connectors.rst
```diff
@@ -0,0 +1,109 @@
...
+.. image:: connector_types.jpg
+
+To actually draw the connectors, additional properties (aside X and Y position) are needed like:
+- style of the line (plain, dotted, dashed, etc.)
+- thickness of the line
+- color of the line
+- presence of arrow heads on both ends
+- style of the arrow heads
+- size of the arrow heads (absolute or relative)
+- color of the arrow heads
```
nyalldawson: I'd strongly suggest using the existing line symbol layer implementation to actually draw the connectors. This would allow much more flexibility for the appearance of connectors, and avoids fragmenting the code with another styled-line-drawing implementation. It also means connectors could benefit from all the existing features in line symbology, eg data defined properties, marker symbols, offsets, etc.

QEP-label-connectors.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Placement of the two ends also need additional properties:
+- position of the ending on the side of the label : up, down, right, left, center, etc.
```
nyalldawson: I also think an exact relative x/y placement should be allowed for the start and end. MapInfo doesn't allow this and only gives positions eg "left of label", "right of label", etc, and it's no-way near flexible enough for precise cartographic placement of leading lines.

QEP-label-connectors.rst
```diff
@@ -0,0 +1,109 @@
...
+lineColor
+headStyle
+headColor
+tailStyle
+tailColor
+isCurved
+void render( point[], renderContext )
+
+In the current code base, the following methods would be modified:
+- QgsPalLabeling::drawLabeling() : extract necessary property values and data-defined values for connectors
```
nyalldawson: This needs to be updated for the new labeling engine

QEP-label-connectors.rst
```diff
@@ -0,0 +1,109 @@
...
+  - for polygons: centroid of the polygon or on the border
+
+Certain types of connectors may also need specific properties like coordinates of the control points for curves (Bézier or B-splines). Coordinates
+of control points can only be data-defined or defined by 'null' which will make the display engine use 'best' coordinates.
+
+Implementation details
+----------------------
+
+Connectors are drawn to the Qt canvas, using the QPainter API. In order to be able to reuse the drawing code in other parts (possibly in a new symbology for lines), some
+classes will be designed.
```
nyalldawson: Again, I'm strongly against having a seperate method to draw these. The curved connector should instead be rendered using the existing symbology classes, and the existing line symbology options expanded where required.

QEP-label-connectors.rst
```diff
@@ -0,0 +1,109 @@
...
+.. image:: connector_properties.png
+
+Each of the property could also be data-defined (not represented here).
+
+From the user perspective, new tools will be available to easily manipulate label connectors. To be consistent with current label manual placement toolboxes,
+new controls will be added to manually place control points for connector types that need them.
+
+Optionaly, it would be nice for the user to be able to edit per-connector properties, i.e. properties that are data-defined via a GUI dialog. This could
+be achieved by reusing the connector properties dialog and connecting it to one particular feature.
+
```
nyalldawson: Again, I like the general direction this is going, with the exceptions noted above. 

Have you thought about automatic creation of a connector when "display a connector" is enabled and the user first moves a label from it's automatic placement? 

Also, what considerations have been given about automatic creation of dynamic connectors if a label is further than a specified tolerance from a feature? Hopefully we'll eventually also see support for that in QGIS, so this work should be done allowing for that possibility in future.

QEP-label-mask.rst
```diff
@@ -0,0 +1,75 @@
...
+Core Changes
+------------
+
+A new type of map layer will be added: QgsMaskLayer inherited from QgsMapLayer. It access a mask geometry that is stored as a global property of the project (through either a QgsProject custom
+property for instance).
+
+Creation of the a new mask layer will be refused if there is already one created (or the existing will be used, if possible).
+
+A new labeling property will be available for each vector layer to use or ignore the mask geometry as an obstacle for the automatic placement of labels.
+
```
nyalldawson: How will the layer be stored? Does this tie in to the proposed new project format and storage of the layer would be in the sqlite project database?

QEP-label-path.rst
```diff
@@ -0,0 +1,71 @@
...
+However for linestrings defined with lots of control points (such as waterways) or with high curvatures, displaying correct
+labels can be either unesthetic or very slow.
+
+This is where using a path slightly different from the feature's geometry could be useful. For instance using a simplified version
+of the linestring could lead to a label faster to display or with smaller curvatures.
+
+We then propose to allow the "curved" placement to use an arbitrary geometry, through a data-defined "label path" property. The data-defined
+property could point to a user-defined column or to an expression.
+
+We also propose to add a "simplify()" function in core expression functions for that purpose.
```
nyalldawson: I've played with this a bit in the past and found that the default Douglas–Peucker simplification method doesn't help at all for curved labeling, as the sharp corners in the linear geometry are preserved. From my tests a better approach is:
- Use Visvalingam’s algorithm to simplify, as that removes sharp corners whilst preserving the overall shape of a line
- Combine smoothing with simplification (see http://qgis.org/api/classQgsGeometry.html#ab2f6d720b51f96686f3152a6dabfe201 ). Note that smoothing increases the number of vertices in a geometry, so it needs to be done after simplifying the geometry.
- Even better would be a "bend simplify" based on the algorithm described in http://web.cs.sunyit.edu/~poissad/projects/Curve/pdf/1998.pdf . It's a more effective method of maintaining the general shape of lines will removing small fluctuations.

QEP-label-path.rst
```diff
@@ -0,0 +1,71 @@
...
+
+This option for placement along a user-defined path would also be useful for other types of features, like points and polygons.
+
+Core Changes
+------------
+
+A new data-defined labeling property of type QgsGeometry is added for the definition of the label path ("labeling/dataDefined/labelPath").
+QgsPalLayerSettings::registerFeature will be modified in consequence: if this new property is set and curved placement is enabled, it will be used as a support.
+
```
nyalldawson: Note that it's not sufficient to just use the alternative geometry when registering a label. Instead, labels needs to be able to register an optional "obstacle geometry" which can differ from the feature geometry. Otherwise the simplified/smoothed line feature would be used for the obstacle detection too, and could result on labels being placed over the rendered symbol for a line.

Note that I've got a nearly-complete implementation of this I hope to merge soon (allowing labels to have different obstacle geometry).

QEP-label-path.rst
```diff
@@ -0,0 +1,71 @@
...
+.. image:: curved_labeling.png
+
+However for linestrings defined with lots of control points (such as waterways) or with high curvatures, displaying correct
+labels can be either unesthetic or very slow.
+
+This is where using a path slightly different from the feature's geometry could be useful. For instance using a simplified version
+of the linestring could lead to a label faster to display or with smaller curvatures.
+
+We then propose to allow the "curved" placement to use an arbitrary geometry, through a data-defined "label path" property. The data-defined
+property could point to a user-defined column or to an expression.
```
nyalldawson: A core change which would be required to accompany this would be to allow access to other geometry columns defined within a layer. Eg for many data formats (Postgis, spatialite) it's possible to have more than one geometry column. These additional geometries are currently just converted to wkb representations and displayed as text fields by QGIS (consequently bogging down use of the attribute table for layers with complex secondary geometries). Support should be added so that additional geometry columns are stored as QgsGeometry objects, and then the label path could be directly bound to the secondary geometry column without the expense of going through the expression engine. This would also allow for the following use case:

- Polygon layer with a secondary line geometry attribute
- Data defined label path is bound to line geometry
- User can add the layer a second time to QGIS, using the line geometry as the main geometry column
- User can then edit the line geometries using the existing digitizing tools, and the labels for the polygon layer will instantly update to reflect any changes to paths.

nyalldawson: Also should probably be noted that when the data defined path geometry is null, feature's geometry will be used instead

nyalldawson:
Great stuff! It's exciting to see this stuff move forward. 

I've made a bunch of notes above reflecting my thoughts on these proposals

QEP-auxiliary-storage.rst
mhugo: Good point

QEP-auxiliary-storage.rst
mhugo: There is no clear timeframe for now. If it happens before the 3.0, backward compatibility will be ensured.

QEP-auxiliary-storage.rst
mhugo: By returning errors to the user when he tries to create new auxiliary fields (and by disabling the automatic creation of such fields). I'll add that.

QEP-auxiliary-storage.rst
mhugo: Yes that what I had in mind. I'll add it to the doc

QEP-auxiliary-storage.rst
mhugo: This section appears after the label path QEP. I was concerned about possible performance issues.
But your remarks are right, thanks.
So ok to remove it or rework it as another (future) QEP for now

QEP-label-connectors.rst
mhugo: I am ok with that

QEP-label-connectors.rst
mhugo: Ok, I'll add it

QEP-label-connectors.rst
mhugo: I like this idea of automatic creation. Adding a connector when the user moves a label is a complement to the automatic creation of aux fields.
The second point would be nice as well.

QEP-label-mask.rst
mhugo: Not sure of the details yet, but yes probably. If the project file is to be reworked, than storing it in a specific (spatialite?) table would be simple.

QEP-label-path.rst
mhugo: Thanks for these references !

QEP-label-path.rst
mhugo: Yes allowing more than one geometry column is an alternative. I am still not sure how big the modifications would be.
Modifying QgsFeature and QgsFeatureIterator to have more than one geometry would probably be sufficient (and providers too).
Something like QgsFeature::geometry( geometryColumn = 0 ) and corresponding methods to get the number of geometry columns ?

mhugo:
@dakcarto @nyalldawson Thanks for your feedbacks ! I'll try to update the documents soon

NathanW2:
QEPs done via issues now. See #27

QEP-auxiliary-storage.rst
haubourg: Hi, no urgent time frame, 3.0 could be the target. the QEP is here to help us clarify the roadmap.

mhugo:
@NathanW2 Ok. I'll also add the 4 other QEPs as issues

NathanW2:
@mhugo would you like me to move you other QEP's or are you happy to do it?  Sorry I didn't read the text to see they are linked in here.

PR #305, opened 2024-10-10
Update README.md

rouault:
We don't want to encourage direct push to master, do we?

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,4 +33,4 @@ QEP numbers will be assigned by [@NathanW2](https://github.com/NathanW2) when is
 
 ### Do I need a QEP for a bug fix?
 
-Nope. Just open a PR or push directly if you have commit rights
+Nope. Just open a PR


nyalldawson (approved):

nyalldawson:
I think we need to rethink how this repo works. The issues approach makes it really hard to see what's accepted vs what's just proposed. I'd be in favour of moving to prs as proposals, accepted as rst for approved proposals. More like the approach gdal uses...

rouault:
> The issues approach makes it really hard to see what's accepted vs what's just proposed. I'd be in favour of moving to prs as proposals, accepted as rst for approved proposals. More like the approach gdal uses...

ah ah yes that was exactly my thoughts. But, without changing too much the process, just making sure that issues in https://github.com/qgis/QGIS-Enhancement-Proposals/issues are consistently labelled with "accepted", "implemented", "rejected", "draft" would be a huge way forward

esnyder-rve:
What about just plain markdown versus rst? Are the "advanced" features of rst needed?

PR #310, opened 2024-12-09
Rework process to use pull requests instead of issues

nyalldawson:
Summary of changes:

- Require pull requests for submission of new QEPs instead of issues
- Document that QEPs which describe policies (such as coding conventions) are MUTABLE, and that proposals can be sent in (via pull requests) to amend and existing QEP. The policies for accepting/rejecting these changes are the same as submission of a new QEP.
- Add note that PRs should be closed when a proposal is rejected -- this way the PR queue will contain ONLY those submissions currently being discussed

### DIFF
--- a/QEP-0-Template.md
+++ b/QEP-0-Template.md
@@ -6,8 +6,6 @@
 
 **Contact** user at domain dot tld
 
-**maintainer** @user
-
 **Version** QGIS X.X
 
 # Summary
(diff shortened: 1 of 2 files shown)


troopa81:
+1

alexbruy:
I'm very much in favor of this approach. Just one question: what we will do with the existing QEPs, will we move them to PRs or leave as is?

+1

nyalldawson:
@alexbruy 

> will we move them to PRs or leave as is?

Well, we have the issue that none of the existing ones were ever really formally approved or voted on, so they are all in a "limbo" state. For this reason I'd lean towards just leaving them as issues and basically start afresh.

PR #314, opened 2024-12-31
QEP 314: Code style and practice guidelines

nyalldawson:
This QEP documents coding standards and conventions used throughout the QGIS code base. Developers are required to follow these standards when submitting code to QGIS.

It is initially intended as a formal description of CURRENT CODE PRACTICES **ONLY**, which can be later modified and revised (in an atomic manner) if we want to change any of these standards. As such, **please refrain from using this initial request as a place to discuss changes to current practice**. Rather, let's keep discussion focused on whether this is an accurate and complete description of our current coding practices. 👍

The end goal here is having a formal policy in place describing all our coding conventions, which will help ease new developers into contributing effectively to QGIS and remove some of the unspoken assumptions surrounding QGIS development.


nyalldawson:
Note: I've arbitrarily restarted the number at 400 here -- this seems right after the restructuring of QEPs to PRs.

qep-400-coding-practice.md
```diff
@@ -0,0 +1,90 @@
...
+
+
+### 3. Language features
+
+- 3.1. For readability and ease of code review, avoid use of ``auto``. The following exceptions are permitted:
+
+  - 3.1.1. ``auto`` should be used for complex types, such as iterators. Eg ``for ( auto it = object.begin(); ...)``
+  
+- 3.2. If ``enums`` are to be used outside of a single .h/.cpp file (or there is a reasonable chance that they will be in future!), they should be placed inside the ``Qgis`` namespace.
+
```
rouault, line 83:
```suggestion

- 3.3. Checking if a pointer is null should be done with ``if ( !ptr )``, omitting explicit comparison with the ``nullptr`` constrant.
```

rouault (approved):

qep-400-coding-practice.md
```diff
@@ -0,0 +1,90 @@
...
+  that method was added. If the method is to be backported to a stable branch, ensure that the ``\since``
+  version correctly describes version at which that method is guaranteed to be accessible. (eg ``\since QGIS 3.34.8``
+  instead of ``\since QGIS 3.34``). Methods introduced in the same version as their class do not require an explicit since annotation.
+- 2.6. Avoid re-documenting overridden methods, unless there is **specific** behavior differences in the override
+  which added extra useful information over the base virtual method documentation.
+- 2.7. Use ``\see otherMethod()`` to link getters and setters (and other related methods)
+- 2.8. The literal values ``true``, ``false`` and ``nullptr`` should be written in documentation as ``TRUE``, ``FALSE`` and ``NULLPTR`` respectively. (Preprocessing macros correctly replace these with ``true``, ``false`` and ``nullptr`` for the c++ documentation, and ``True``, ``False`` and ``None`` for the PyQGIS documentation)
+- 2.9. Include explicit details of memory ownership in the documentation. Eg "ownership of the symbol is transferred to the renderer", "caller takes ownership of thr returned object".
```
DelazJ, line 73:
```suggestion
- 2.9. Include explicit details of memory ownership in the documentation. Eg "ownership of the symbol is transferred to the renderer", "caller takes ownership of the returned object".
```

qep-400-coding-practice.md
DelazJ: @rouault s/constrant/constant?

uclaros:
> Note: I've arbitrarily restarted the number at 400 here -- this seems right after the restructuring of QEPs to PRs.

Do we actually need to have a QEP numbering system that does not match the PR/issue number? It seems to me like a potential point of confusion and a burden to keep track of!
314 (being 100 * π) is an easy to remember number for the first PR based QEP :)

nyalldawson:
@uclaros 
> Do we actually need to have a QEP numbering system that does not match the PR/issue number? 

I can certainly change it, but keep in mind that we'll end up with big gaps in the numbers that way, since issues and pr numbering is shared but only prs will be QEPs. It's also possible now to have multiple prs target the same QEP (eg a future change to these specifications is modifying the same QEP from a different pr number). Given this, I don't see any benefit in forcing the numbering to match the first pr creating a new QEP.

uclaros:
> I can certainly change it, but keep in mind that we'll end up with big gaps in the numbers that way, since issues and pr numbering is shared but only prs will be QEPs. It's also possible now to have multiple prs target the same QEP (eg a future change to these specifications is modifying the same QEP from a different pr number).

Do we mind if there are gaps? We were using the issue number until now so gaps were inevitable.
Regarding multiple PRs for a single QEP we can have a `QEP` label applied to the main PR and keep referencing it by its number.

I'd rather have gaps in the numbering than having to search the closed PRs list for the last QEP number to increment, unless there's some simple way to automate this.

m-kuhn (approved):

elpaso:
@nyalldawson looks goo to me, thank you for taking care of this.

Should we also mention the use of (possibly redundant but explicit) `std::as_const` in for loops?

troopa81 (approved):
Should we also include something about using the new settings API  instead of the raw string Qt api?

troopa81:
@uclaros Maybe that should come in a separate PR. Do you mind if I merge this one and we iterate later on it?

nyalldawson:
@troopa81 

Please don't merge -- I want this initial version to be complete.

troopa81:
> Please don't merge -- I want this initial version to be complete.

OK

m-kuhn:
@nyalldawson I have converted to draft, with the approvals in place already it will be easier to monitor, when you are happy with it.

qep-400-coding-practice.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Coding Guidelines
+
+### 1. Naming
+
+- 1.1. Variable and function naming should be camel case. Generally Qt's approach to capitalising only the first letter in acronyms is used (eg ``QgsGpsDetector``, not ``QgsGPSDetector``), with the exception of ``3D`` (eg ``Qgs3DMapCanvas``).
+
+  - 1.1.1. Local variables should have no name prefix (eg do not prefix variable names with ``my``)
+  - 1.1.2. [Hungarian notation](https://en.m.wikipedia.org/wiki/Hungarian_notation) is NOT used, except in some exceptional circumstances (eg variable names used with the gdal c API)
+  - 1.1.3. Member variables should have a ``m`` prefix (eg ``mLineWidth``)
```
troopa81, line 20: Does it also apply to struct? It doesn't seem consistent over the code base
- [with](https://github.com/qgis/QGIS/blob/64e2a010199db7a655aab70ae968966e550efc4d/src/core/expression/qgsexpression.h#L92)
- [without](https://github.com/qgis/QGIS/blob/64e2a010199db7a655aab70ae968966e550efc4d/src/core/effects/qgsimageoperation.h#L221)

Maybe it's worth mentioning here wether it applies to class AND struct or only to the first one

3nids:
One small question regarding the process and scope of QEPs.

To me a QEP should be definitive once merged/accepted.

For this one, it looks like it will be a continuous evolution. So is this the correct place?

For code style, we already have this: https://github.com/qgis/QGIS-Documentation/blob/master/docs/developers_guide/codingstandards.rst which sounds more appropriate than a QEP.

nyalldawson:
@uclaros 
Renamed to 314.

nyalldawson:
@elpaso 

> Should we also mention the use of (possibly redundant but explicit) std::as_const in for loops?

Ok, I've added that for explicitly non const containers only. I don't think we currently have agreement on whether this should be required for already const containers, so if you want that then let's start that change discussion separately after this QEP is formalised.

nyalldawson:
@troopa81 
> Should we also include something about using the new settings API instead of the raw string Qt api?

No, IMO that one would belong in a formal policy regarding settings usage. (Ideally with code examples for both C++ and python)

nyalldawson:
@3nids 
> One small question regarding the process and scope of QEPs.
> To me a QEP should be definitive once merged/accepted.
> For this one, it looks like it will be a continuous evolution. So is this the correct place?

Yes, that's correct. Since https://github.com/qgis/QGIS-Enhancement-Proposals/pull/310 QEPs can now be mutable, exactly so that we have a way of formally handling mutable QEPs just like this one.

There's a mix of needs for QEPs. We have short lived project based ones like https://github.com/qgis/QGIS-Enhancement-Proposals/pull/318 which likely will remaining unchanged forever after the initial acceptance. Then we have "living" specs like this one, or eg one for backport policies, pull request standards, API stability requirements, etc. These ones are inherently mutable and trying to handle them via seperate independent QEPs where every change is a new document is impractical. The new approach allows for a single point of truth for living specifications like this.

> For code style, we already have this: https://github.com/qgis/QGIS-Documentation/blob/master/docs/developers_guide/codingstandards.rst which sounds more appropriate than a QEP.

That page should be removed and changed to a link to this formal specification after merging.

nyalldawson:
@uclaros 

> Separate qgis from qt includes
> Include qt as #include <QMap> and not #include "qmap.h"
> Use QStringLiteral for untranslated literals, QLatin1String for comparisons
> Don't use qDebug(), use QgsDebugError/QgsDebugMsgLevel 

Thanks, I've added these

> Include moc file for qobjects
> Header/license for each file

These are already explicitly covered by tests on the CI, I'll omit them as a result.

> Prefer forward declaration if possible instead of including from within header files
methods that are actually const should be marked const

These fall under "best practice modern C++" , so I don't want to explicitly add them here. (And the second is already covered by cppcheck on CI)

qep-400-coding-practice.md
nyalldawson: Indeed, there's no consistency here. I'll omit any formal specification for now until we decide on one.

nyalldawson:
Ok, all comments addressed. How are we looking now?

3nids (approved):

elpaso:
> @elpaso
> 
> > Should we also mention the use of (possibly redundant but explicit) std::as_const in for loops?
> 
> Ok, I've added that for explicitly non const containers only. I don't think we currently have agreement on whether this should be required for already const containers, so if you want that then let's start that change discussion separately after this QEP is formalised.

IIRC it was @m-kuhn back in the days who corrected my code when I didn't use as_const over an already const container. I have always thought that it was pointless but also harmless (it would be a no-op in that case), I'm fine with not requiring as_const over an already const container.

elpaso (approved):

m-kuhn:
I do think that reviewer time and focus is not best invested to check if a variable over which we loop is already const in a given context, but I am open how we handle this. I would keep this discussion separate and not block this, given there is no agreement yet.

nyalldawson:
Declaring as accepted with approval from @troopa81 @rouault @m-kuhn @elpaso @3nids

nyalldawson:
@3nids @DelazJ Documentation update was submitted as https://github.com/qgis/QGIS-Documentation/pull/9581

PR #318, opened 2025-01-07
QEP: Overview of widget types while configuring forms

gacarrillor:
QEP to add some UI/UX enhancements when configuring forms and/or wanting a glimpse on what's already configured.

Since the QEP process is being refactored, we're open to adapt this PR to any given format :)

### DIFF
--- a/qep-401-overview-widget-types.md
+++ b/qep-401-overview-widget-types.md
@@ -1,0 +1,74 @@
+# QGIS Enhancement: Overview of widget types, constraints and default values
+
+**Date** 2025/01/07
+
+**Author** Germán Carrillo (@gacarrillor), Mathieu Pellerin (@nirvn)
+
+**Contact** german at opengis dot ch, mathieu at opengis dot ch
+
+**Version** QGIS 3.4X.X+
+
+# Summary
+
+
+In the vector layer properties, QGIS allows users to drag available layer fields from a field panel and drop them in a form layout panel. Once there, each form layout item has a widget type assigned with a default configuration. These widget types are currently only shown individually in a third panel, where users can change both the widget type and its configuration.
+
+![Screenshot from 2024-11-24 22-22-18](https://github.com/user-attachments/assets/47cee69d-6ff2-4f50-9069-eab44bb66757)
+
+When a layer has a relatively large number of fields, there is a need to have an overview of widget types, rather than just seeing them one by one. This can be easily extended to other field settings like constraints and default values.
+
+We propose to tackle this via visual hints in field and form item panels, as well as bringing handy stuff to improve user experience.
+
+## Proposed Solution
+
+Switch from current item-based tree widget to model-based tree view to ease maintenance and gain flexibility.
+
+Use the form layout item’s decoration role (i.e., item’s icon) to show its type of widget. This is appropriate because all items will have one and only one type of widget.
+
+Use right-aligned indicator icons (like those that are used in the QGIS layer tree for filters and missing data sources, among others) for constraints and default values. The indicators will only be shown when constraints and/or default values are present, and details about specific configurations will be displayed via tooltips.
+
+This work includes the graphic design of almost 20 icons for widget types, as well as a couple of icons for indicators of constraints and default values.
+
+
+#### Other UI/UX enhancements
+
+ + **Filter/search box**, taking advantage of the model-based tree view.
+
+ + **Form preview**, to give users the option to see what they'll get easily and in the right place, i.e., without having to follow additional steps like closing Layer properties, enabling editing and adding a feature just to see what they've just configured in action. The preview will be opened in a new window, parented to the Vector Layer Properties dialog.
+
+ + In the form layout panel, users will have the possibility to **switch between aliases and item names**, depending on their needs. This is particularly handy to easily find form layout items as they will be displayed in the generated input form, which may be cumbersome nowadays.
+
+ + The widget configuration panel (i.e., the right-hand side panel) currently shows display, field and widget associated configurations in a single page. We'll separate them in **three tabs**:
+
+    + Display: display group;
+    + Field: general group, constraints, default values;
+    + Widget: widget type.
+
+     To ease association, the Field tab could get the focus when a user selects an available item in the left-hand side panel, whereas the Widget tab would get the focus when a user selects an item in the form layout panel (i.e., the one in the middle).
+
+## Deliverables
+
+Pull request(s) in QGIS repo implementing the enhancements listed and described above, including the new set of icons.
+
+### Affected Files
+
+`src/gui/attributeformconfig/*`
+`src/gui/vector/qgsattributesformproperties.cpp`
+`src/gui/vector/qgsattributesformproperties.h`
+...
+
+## Risks
+
+None.
+
+## Performance Implications
+
+Not applicable, since the proposal deals with displaying already available data from items.
+
+## Backwards Compatibility
+
+The proposal will also work with projects from older QGIS versions.
+
+## Issue Tracking ID(s)
+
+https://github.com/qgis/QGIS/issues/58973
--- a/qep-401-overview-widget-types.md
+++ b/qep-401-overview-widget-types.md
@@ -33,18 +33,8 @@ This work includes the graphic design of almost 20 icons for widget types, as we
 #### Other UI/UX enhancements
 
  + **Filter/search box**, taking advantage of the model-based tree view.
-
  + **Form preview**, to give users the option to see what they'll get easily and in the right place, i.e., without having to follow additional steps like closing Layer properties, enabling editing and adding a feature just to see what they've just configured in action. The preview will be opened in a new window, parented to the Vector Layer Properties dialog.
-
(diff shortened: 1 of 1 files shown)


nyalldawson:
Regarding "form preview" -- how will this be coded? Will you add API to allow showing forms with a specific edit form configuration, or will the changes be applied to the layer first? I'd much rather the former approach, that API sounds like it would be quite useful (and would eg pave the way for multiple form configuration for a layer)

uclaros (commented):
Nice to see some <3 for the form designer!

It would also probably be good to have the extra goodies (widget type, constraint/default icons) for _Autogenerate_ forms where the _Form Layout_ view is not available.

gacarrillor:
> Regarding "form preview" -- how will this be coded? Will you add API to allow showing forms with a specific edit form configuration, or will the changes be applied to the layer first? I'd much rather the former approach, that API sounds like it would be quite useful (and would eg pave the way for multiple form configuration for a layer)

The new API would be the way to go in this case.

qep-401-overview-widget-types.md
```diff
@@ -0,0 +1,74 @@
...
+
+When a layer has a relatively large number of fields, there is a need to have an overview of widget types, rather than just seeing them one by one. This can be easily extended to other field settings like constraints and default values.
+
+We propose to tackle this via visual hints in field and form item panels, as well as bringing handy stuff to improve user experience.
+
+## Proposed Solution
+
+Switch from current item-based tree widget to model-based tree view to ease maintenance and gain flexibility.
+
+Use the form layout item’s decoration role (i.e., item’s icon) to show its type of widget. This is appropriate because all items will have one and only one type of widget.
```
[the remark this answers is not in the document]
gacarrillor: Yes, it would be a challenge to create the set of icons for this. Fortunately, @nirvn would be in charge.

We've thought about additional columns as well, but for the moment would like to avoid them if possible due to the lack of space in that part of the UI.

Anyway, lets see what others think about it.

qep-401-overview-widget-types.md
```diff
@@ -0,0 +1,74 @@
...
+
+ + **Form preview**, to give users the option to see what they'll get easily and in the right place, i.e., without having to follow additional steps like closing Layer properties, enabling editing and adding a feature just to see what they've just configured in action. The preview will be opened in a new window, parented to the Vector Layer Properties dialog.
+
+ + In the form layout panel, users will have the possibility to **switch between aliases and item names**, depending on their needs. This is particularly handy to easily find form layout items as they will be displayed in the generated input form, which may be cumbersome nowadays.
+
+ + The widget configuration panel (i.e., the right-hand side panel) currently shows display, field and widget associated configurations in a single page. We'll separate them in **three tabs**:
+
+    + Display: display group;
+    + Field: general group, constraints, default values;
+    + Widget: widget type.
```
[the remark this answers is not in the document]
nyalldawson: I'm also not a fan of including tabs in this panel. That's going to lead to a very messy UI when opening the layer properties window (vertical tabs on the left, triple panel layout in the centre, with tab bar on the right...  😖)

qep-401-overview-widget-types.md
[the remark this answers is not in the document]
troopa81: I'd prefer the icon over text in an additional column which would take too many place. Still, there is a challenge to make clean and recognizable icons.

qep-401-overview-widget-types.md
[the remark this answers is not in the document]
troopa81: -1 for tabs, and collapsible group box are good enough to show only what user wants to focus on.

qep-401-overview-widget-types.md
```diff
@@ -0,0 +1,74 @@
...
+Use right-aligned indicator icons (like those that are used in the QGIS layer tree for filters and missing data sources, among others) for constraints and default values. The indicators will only be shown when constraints and/or default values are present, and details about specific configurations will be displayed via tooltips.
+
+This work includes the graphic design of almost 20 icons for widget types, as well as a couple of icons for indicators of constraints and default values.
+
+
+#### Other UI/UX enhancements
+
+ + **Filter/search box**, taking advantage of the model-based tree view.
+
+ + **Form preview**, to give users the option to see what they'll get easily and in the right place, i.e., without having to follow additional steps like closing Layer properties, enabling editing and adding a feature just to see what they've just configured in action. The preview will be opened in a new window, parented to the Vector Layer Properties dialog.
```
troopa81, line 37: Instead of having another windows preview, could we not just update the real form when user clicks on *Apply* button ? It's actually the case when you change the widget, but not when we change the alias for instance.

If it was working for any parameter you can modify, it would solve the preview issue in an even better way, because it would be populated with real data.

troopa81:
Nice proposal @gacarrillor :+1:

gacarrillor:
> Nice proposal @gacarrillor 👍

Thank you @troopa81.
Credit is due to @3nids, @nirvn and @m-kuhn, who shaped it as well :)

qep-401-overview-widget-types.md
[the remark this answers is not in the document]
3nids: let's leave this aside then.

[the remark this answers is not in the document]
gacarrillor: Thanks for the feedback, applied in https://github.com/qgis/QGIS-Enhancement-Proposals/pull/318/commits/08eff5a604c0d95cfa12dd071fb1a208589f13b5

qep-401-overview-widget-types.md
gacarrillor: We haven't really aimed at changing existing QGIS behavior on opened forms. The form preview will only be shown as a child window of the Vector Layer Properties dialog. Users could preview the configured form without leaving that dialog, i.e., without having to open the attribute table's form view or to add a new geometry.

troopa81: I understand but it's just yet another window (and more code) when there is actually one existing, already populated with relevant data so you can better look about what it really look like.

Do you plan to populate this preview window with a feature like what we are doing for expression evaluation for instance, or it's more like the Qt designer empty preview?

m-kuhn: Fair question about how the form will be populated with data.

Regarding the proposal to reuse the "real" form, I wonder if that's a good experience. The layer properties dialog (where widgets/forms are configured) is a modal dialog, so it won't be possible to open a form through the main window and already open forms will be stacked behind the layer properties and not interactive. I am not sure we can realistically reuse these.

troopa81: Fair point too... I tested it before posting my comment but I was actually using dual screen.

I like current behavior with dual screen though
widgets

here we see that form is not refresh when alias is modified, but only when widget type is.

We could also do like we do with map symbology and display widget editor beside the attribute form when hitting F7 and refresh the form as soon as user modify a parameter.

gacarrillor: > Do you plan to populate this preview window with a feature like what we are doing for expression evaluation for instance, or it's more like the Qt designer empty preview?

I guess I always had this Qt-designer's empty preview in mind (i.e., that would already be great), but adding sample data to the preview in the same way that it is done for previewing expressions results sounds awesome. 

Of course, we won't have a `Save` button available there, but the rest could look like a real input form. That is, widgets wouldn't be set to read-only mode (unless the widget config says so), but any changes to the sample data wouldn't be applied.

gacarrillor:
> In ideal world, we should have a separate window (similar to print composer or model builder) to handle forms and tables. It would allow users to create relations graphically, select widgets, preview forms, create a new table using widgets directly (similar to Google forms), etc. The current state under layer properties is limiting in a long term and is not scaling well with all the great functionalities being added constantly to forms. I understand your scope is limited, so the this might be out of consideration.

Nonetheless, those are good guidelines to keep in mind, thank you!

qep-401-overview-widget-types.md
gacarrillor: > Do you plan to populate this preview window with a feature like what we are doing for expression evaluation for instance, or it's more like the Qt designer empty preview?

Suggestion applied in https://github.com/qgis/QGIS-Enhancement-Proposals/pull/318/commits/bcd86d40a150575c78556a434ad4ce42ca58c16c, thanks!

> We could also do like we do with map symbology and display widget editor beside the attribute form when hitting F7 and refresh the form as soon as user modify a parameter.

With respect to this, it looks to be in-line with what @saberraz suggested for future iterations.
We'd consider moving the UI to a docked panel a bit out of scope for now.

nyalldawson:
@gacarrillor what's the status here? Is this ready to be put to vote/approval?

gacarrillor:
> @gacarrillor what's the status here? Is this ready to be put to vote/approval?

Thanks for raising this. We planned to send the communication this week, so let's officially start the voting process. I just sent the message to the dev-list.

nyalldawson:
+1

elpaso:
+1

nyalldawson:
Declaring as accepted with votes by @nyalldawson , @elpaso

PR #319, opened 2025-01-22
QEP 314: Relax prohibition against "auto"

nyalldawson:
Allow use of ``auto`` for variable types when the variable type is explicitly stated during its initialization.

Specifically:

``auto`` may be used for variable types if the type is explicit during variable initialization. Eg

```
// allowed, as the QgsPoint type is explicit during initialization:
auto pointObject = QgsPoint( 3, 4 );
// allowed, as the std::unique_ptr< QgsPoint >, std::shared_ptr< QgsPoint > types are explicit during initialization:
auto pointUniquePointer = std::make_unique< QgsPoint >( 3, 4 );
auto pointSharedPointer = std::make_shared< QgsPoint >( 3, 4 );

// NOT allowed, the argument types for the std::tuple are not explicit:
auto myTuple = std::make_tuple( 0, 5 );
```

### DIFF
--- a/qep-314-coding-style.md
+++ b/qep-314-coding-style.md
@@ -77,7 +77,7 @@ Notes:
 
 - 3.1. For readability and ease of code review, avoid use of ``auto``. The following exceptions are permitted:
 
-  - 3.1.1. ``auto`` should be used for complex types, such as iterators and lambda functions. Eg ``for ( auto it = object.begin(); ...)``
+  - 3.1.1. ``auto`` should be used for complex types, such as iterators and lambda functions. Eg ``for ( auto it = object.constBegin(); ...)``
   - 3.1.2. ``auto`` may be used for ``std::unique_ptr`` and ``std::shared_ptr`` types if the pointer type is explicit during variable initialization. Eg
 
 ```
--- a/qep-314-coding-style.md
+++ b/qep-314-coding-style.md
@@ -77,18 +77,16 @@ Notes:
 
 - 3.1. For readability and ease of code review, avoid use of ``auto``. The following exceptions are permitted:
 
-  - 3.1.1. ``auto`` should be used for complex types, such as iterators. Eg ``for ( auto it = object.begin(); ...)``
-  - 3.1.2. ``auto`` may be used for variable types if the type is explicit during variable initialization. Eg
+  - 3.1.1. ``auto`` should be used for complex types, such as iterators and lambda functions. Eg ``for ( auto it = object.begin(); ...)``
+  - 3.1.2. ``auto`` may be used for ``std::unique_ptr`` and ``std::shared_ptr`` types if the pointer type is explicit during variable initialization. Eg
 
 ```
-// allowed, as the QgsPoint type is explicit during initialization:
-auto pointObject = QgsPoint( 3, 4 );
 // allowed, as the std::unique_ptr< QgsPoint >, std::shared_ptr< QgsPoint > types are explicit during initialization:
 auto pointUniquePointer = std::make_unique< QgsPoint >( 3, 4 );
 auto pointSharedPointer = std::make_shared< QgsPoint >( 3, 4 );
 
-// NOT allowed, the argument types for the std::tuple are not explicit:
-auto myTuple = std::make_tuple( 0, 5 );
+// NOT allowed, the unique_ptr type is not explicit:
+auto myUniquePtr = createObjectFunction( 0, 5 );
 ```
 
 - 3.2. If ``enums`` are to be used outside of a single .h/.cpp file (or there is a reasonable chance that they will be in future!), they should be placed inside the ``Qgis`` namespace.
--- a/qep-314-coding-style.md
+++ b/qep-314-coding-style.md
@@ -78,7 +78,19 @@ Notes:
 - 3.1. For readability and ease of code review, avoid use of ``auto``. The following exceptions are permitted:
 
   - 3.1.1. ``auto`` should be used for complex types, such as iterators. Eg ``for ( auto it = object.begin(); ...)``
-  
+  - 3.1.2. ``auto`` may be used for variable types if the type is explicit during variable initialization. Eg
+
+```
+// allowed, as the QgsPoint type is explicit during initialization:
+auto pointObject = QgsPoint( 3, 4 );
+// allowed, as the std::unique_ptr< QgsPoint >, std::shared_ptr< QgsPoint > types are explicit during initialization:
+auto pointUniquePointer = std::make_unique< QgsPoint >( 3, 4 );
+auto pointSharedPointer = std::make_shared< QgsPoint >( 3, 4 );
+
+// NOT allowed, the argument types for the std::tuple are not explicit:
+auto myTuple = std::make_tuple( 0, 5 );
+```
+
 - 3.2. If ``enums`` are to be used outside of a single .h/.cpp file (or there is a reasonable chance that they will be in future!), they should be placed inside the ``Qgis`` namespace.
 
 - 3.3. Checking if a pointer is null should be done with ``if ( !ptr )`` or ``if ( ptr )`` alone, omitting explicit comparison with the ``nullptr`` constant.


nyalldawson:
FYI @rouault

qep-314-coding-style.md
```diff
@@ -78,7 +78,19 @@ Notes:
...
+
+```
+// allowed, as the QgsPoint type is explicit during initialization:
+auto pointObject = QgsPoint( 3, 4 );
+// allowed, as the std::unique_ptr< QgsPoint >, std::shared_ptr< QgsPoint > types are explicit during initialization:
+auto pointUniquePointer = std::make_unique< QgsPoint >( 3, 4 );
+auto pointSharedPointer = std::make_shared< QgsPoint >( 3, 4 );
+
+// NOT allowed, the argument types for the std::tuple are not explicit:
+auto myTuple = std::make_tuple( 0, 5 );
```
rouault, line 91: I think the best practice in this kind of document is to provide, when possible, the recommended/mandatory alternative when something is disallowed. So maybe the following? (Not totally sure if striking will work)

```suggestion
// NOT allowed, the argument types for the std::tuple are not explicit:
<s>auto myTuple = std::make_tuple( 0, 5 );</s>
// rather do
auto myTuple = std::tuple<int, int>( 0 , 5 )
```

m-kuhn:
What is the motivation for this?
While it's less typing for the author it's more brainwork for readers.

qep-314-coding-style.md
```diff
@@ -78,7 +78,21 @@ Notes:
 - 3.1. For readability and ease of code review, avoid use of ``auto``. The following exceptions are permitted:
 
   - 3.1.1. ``auto`` should be used for complex types, such as iterators. Eg ``for ( auto it = object.begin(); ...)``
-  
+  - 3.1.2. ``auto`` may be used for variable types if the type is explicit during variable initialization. Eg
+
+```
+// allowed, as the QgsPoint type is explicit during initialization:
+auto pointObject = QgsPoint( 3, 4 );
```
wonder-sk, line 85: Is there any reason we would want to write this instead of `QgsPoint pointObject( 3, 4 );` ? The notation with `auto` is longer, and may involve extra copy of the temporary object?

nyalldawson:
> What is the motivation for this?
While it's less typing for the author it's more brainwork for readers.

It's to make super verbose lines like these less painful to write:

    std::unique_ptr< QgsProcessingParameter > extentParam = std::make_unique< QgsProcessingParameter >( .... ); 

in this case it's still immediately clear to reviewers/readers if it's written like:

    auto extentParam = std::make_unique< QgsProcessingParameter >( .... );

qep-314-coding-style.md
nyalldawson: @wonder-sk 

> Is there any reason we would want to write this instead of QgsPoint pointObject( 3, 4 ); ? The notation with auto is longer, and may involve extra copy of the temporary object?

That's a good point. Should we reword this change to only allow for some explicit types like std::unique_ptr / std::shared_ptr ?

wonder-sk: > Should we reword this change to only allow for some explicit types like std::unique_ptr / std::shared_ptr ?

yeah I think that would be good...

nirvn:
Being allergic to auto, +1 to restricting this relaxation to only std::unique_ptr / std::shared_ptr as suggested above.

elpaso:
Being an auto worshiper, +1 to relaxing.

nirvn:
@elpaso , heh, that had my laugh out loud :)

troopa81:
not a great fan of auto so +1 to restricting to only std::unique_ptr / std::shared_ptr.

dvdkon:
Now I feel like I have to voice my opinion too :) I like `auto` and would prefer allowing it instead of any non-primitive types, especially when the rough type is clear from context.  
For example, I feel that `QgsPointCloudIndex index = mLayer->index();` can be shortened to `auto index = mLayer->index();` without any impact to readability.

By the way, thanks @nyalldawson for these coding style QEPs. A comprehensive coding standard was something I sorely missed when starting with QGIS development.

nyalldawson:
@dvdkon 

> For example, I feel that QgsPointCloudIndex index = mLayer->index(); can be shortened to auto index = mLayer->index(); without any impact to readability.

My issue with that is that it requires an IDE to actually determine the object type. Eg in this case "index()" could quite validly return a QgsSpatialIndex, a QgsKdBushSpatialIndex, or something else entirely (a QHash? A QMap?). In short, it makes pull request reviews orders of magnitude more difficult.

m-kuhn:
+1 for non trivial sounds good. Are iterators listed as acceptable somewhere already?

dvdkon:
Small note: If `auto` is going to be restricted to some exhaustive list of cases, I think initialisation with `new` should also be included.  
It's not nice or modern, but it's still used in many parts of the codebase.

m-kuhn:
> Small note: If `auto` is going to be restricted to some exhaustive list of cases, I think initialisation with `new` should also be included. It's not nice or modern, but it's still used in many parts of the codebase

I think this is worth discussing in a different proposal (to keep this one focussed), there are a few considerations to be made but in general it's a desirable direction IMHO.

rouault:
> I think initialisation with `new` should also be included.

generally speaking, we should avoid using new as much as possible and favor smart pointers (std::make_unique, etc) wherepossible

dvdkon:
> generally speaking, we should avoid using new as much as possible and favor smart pointers (std::make_unique, etc) where possible

I definitely agree, but it's used in the codebase today and it's not always possible to sanely replace it with a smart pointer.

In the future, I'd love to see some discussion on how we can further move away from raw pointers (the main blockers right now are Qt and SIP AFAIK), but for now I just don't want to have to keep writing `LongClassName *longObjectName = new LongClassName;'.

nyalldawson:
> I think this is worth discussing in a different proposal (to keep this one focused)

+1.

@dvdkon would you like to create that QEP?

nyalldawson:
Ok, based on feedback this is now restricted to ``std::unique_ptr`` and ``std::shared_ptr`` only.

qep-314-coding-style.md
```diff
@@ -77,8 +77,18 @@ Notes:
 
 - 3.1. For readability and ease of code review, avoid use of ``auto``. The following exceptions are permitted:
 
-  - 3.1.1. ``auto`` should be used for complex types, such as iterators. Eg ``for ( auto it = object.begin(); ...)``
-  
+  - 3.1.1. ``auto`` should be used for complex types, such as iterators and lambda functions. Eg ``for ( auto it = object.begin(); ...)``
```
m-kuhn, line 80:
```suggestion
  - 3.1.1. ``auto`` should be used for complex types, such as iterators and lambda functions. Eg ``for ( auto it = object.constBegin(); ...)``
```

Let's be conservative here

nyalldawson:
This is ready for voting now

rouault:
+1 Even

nirvn:
+1

alexbruy:
+1

m-kuhn:
+1

troopa81:
+1

elpaso:
+1

nyalldawson:
Declaring as passed with approval from @nirvn, @rouault, @alexbruy, @m-kuhn, @troopa81 and @elpaso

uclaros:
Late to the party, but I wonder what your sentiments would be for the cases of type casting, eg:
```cpp
const QgsVectorLayerUndoPassthroughCommandChangeGeometry *merge = dynamic_cast<const QgsVectorLayerUndoPassthroughCommandChangeGeometry *>( other );

```

nyalldawson:
@uclaros I've personally no issue with that! Would you mind opening a request and kick starting the discussion on it?

nyalldawson:
@uclaros 

>Late to the party, but I wonder what your sentiments would be for the cases of type casting, eg:

See https://github.com/qgis/QGIS-Enhancement-Proposals/pull/330

Issue #151, opened 2019-07-24
Support SAP HANA databases in QGIS


stefanuhrig:
# QGIS Enhancement: Support SAP HANA databases in QGIS

**Date** 2019/07/24

**Author** Stefan Uhrig (@stefanuhrig)

**Contact** stefan.uhrig at sap dot com

**maintainer** @mrylov

**Version** QGIS 3.X

# Summary

[SAP HANA](https://www.sap.com/products/hana.html) is an in-memory database with an [OGC-compliant](http://www.opengeospatial.org/resource/products/details/?pid=1303) [spatial engine](https://help.sap.com/viewer/cbbbfc20871e4559abfd45a78ad58c02/2.0.04/en-US/e1c934157bd14021a3b43b5822b2cbe9.html). A free community edition of SAP HANA is available [here](https://www.sap.com/cmp/td/sap-hana-express-edition.html).

SAP's customer base is in quite some demand for HANA support in QGIS. Most of those customers are already using QGIS, mostly with PostGIS, Oracle and/or MSSQL. They expect that they can work with HANA in QGIS in the same way they work with these databases.

We would like to add HANA support to QGIS. Additionally, we would like to ensure that QGIS users have the same experience when working with HANA as they have working with other databases.

## Proposed Solution

We would like to add a new provider for SAP HANA to QGIS. We have prepared pull request qgis/QGIS#30734 for that.

Some people in the community have already raised concerns about adding a new provider to the QGIS repository and suggested alternatives like providing HANA support as plug-in or accessing HANA via GDAL.

However, we are concerned that we cannot achieve the same kind of integration into QGIS with these alternative approaches as we can by adding a new data provider to QGIS. The experience for our customers will then be different from working with other databases in QGIS. This will likely lead to limited acceptance and unsatisfied customers.

### Affected Files

See qgis/QGIS#30734 for a list of files.

## Backwards Compatibility

We don't expect any backward compatibility issues.

pcav:
I would agree on that, provided there is a long term commitment by at least one developer to keep the provider up to date. Our pas experiences were problematic, as mentioned.
BTW, it would be preferable to also include authentication.

nirvn:
An alternative approach here would be to seize this opportunity to involve the relevant developers to insure serving a provider in the form of a plugin achieves the same level of integration as a core provider. @wonder-sk probably has more to say on that front.

I can't confirm it's currently the case but I'm very certain it should be :)

[...]

PR #326, opened 2025-02-11
Add new QEP-0-Template.md

DMZHY:
Add QEP-0-Template.md and its images.


troopa81:
Could you change the QEP name so it state explicitly what this QEP is about.

troopa81:
Thank you for your proposal @DMZHY and your interest in QGIS.

I have to admit I'm not a great fan about integrating more and more proprietary database in QGIS core. We already have this discussion with SAP [Hana](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/151) support and we accepted to integrate it into core. There is also already Oracle and MSSQL. So why not Dameng?

Did you consider writing a python plugin? We did it for [QDuckDb](https://plugins.qgis.org/plugins/qduckdb/) (which is OpenSource). It works well (There is, I have to admit, some performance implication that might require optimization on core side).

rouault:
> I have to admit I'm not a great fan about integrating more and more proprietary database in QGIS core

Speaking only for myself:

I believe one of our (= QGIS community) concern is that those type of contributions tend to be extremely focused to the provider that the company contributes to (in the best case, they maintain it. in the worse cases, the end up abandoning it, like the DB2 provider we ended up removing), and not participate actively to the rest of the general life of the project. So the project ends up with big chunks of code that are only of use for a very small subset of its users (or at least the part of the users the active contributors usually interact with. I'm well aware that China is a big country and the potential number of users of this database could be big), but that become a liability for the rest of the contributors that have to deal with them on a daily basis and do lots of micro-maintenance tasks on them each time a small or big evolutions of the provider API is done.

My recommendation would be that before pushing a big "present" to QGIS you start by finding your place in the contributing community by helping addressing known issues, small new features of general interest, participating in code reviews, etc. That is showing you care about QGIS by itself, and not just about increasing the market share of your product. That's an on-boarding process that usually takes months.

nyalldawson:
@DMZHY 

A couple of notes:

- you need to copy the QEP-0 template file and then modify the copy, not replace the existing content
- The proposal is missing technical details -- in its current form all your saying is basically "the same as postgres, but for Dameng". Instead, you should focus on what the technical consequences of this proposal are. Does it add extra dependencies to QGIS? Will it always be compiled, or conditionally based on other libraries? How well packaged are those libraries? What CI setup will you implement to test the provider? Which parts of the Dameng spatial database DON'T fit within the QGIS vector data provider model and what trade-offs do you need to make?

And then, as @rouault has pointed out, you need to also describe your plans for maintaining the provider. Who is responsible for bug fixing? Who will maintain the CI setup? Will you be financially sponsoring the QGIS project to help offset the added maintenance burden of adding a new provider?

DMZHY:
Dear QGIS Development Team,

First of all, thank you for your attention and feedback on our enhancement proposal for QGIS adaptation to Dameng Database.

I understand your concerns about integrating more proprietary databases into the QGIS core code, and I fully agree that as QGIS core functionality continues to expand, greater emphasis should be placed on maintaining the universality and sustainability of the code. Our company places great emphasis on the development of spatial databases, and currently, Dameng Database has a dedicated R&D team focused on spatial data. As QGIS becomes widely adopted in China, many QGIS users in the market have requested support for Dameng Database. Therefore, our R&D team has undertaken adaptation work for QGIS. Regarding the code submission for QGIS, we commit to providing ongoing maintenance and support to ensure maintenance and optimization of the adaptation code in the long run.

In addition to database adaptation, our R&D team contributes to various aspects of QGIS open source. We will actively participate in issue reporting, bug fixes, and addressing community issues within the QGIS open-source community, striving to contribute to the development of QGIS. At the same time, we have completed corresponding adaptation work on other GIS platforms such as ArcGIS and SuperMap. Our company has also provided feedback on implementation issues in GIS open-source communities such as GEOS, CGAL, and GDAL.

We hope to have the opportunity to contribute our adaptation code to the QGIS core code. Of course, if it currently does not meet the integration requirements for certain reasons, we fully understand and support that decision. Once this proposal is accepted into the core code for QGIS, we will be sure to modify the copy file name in the format of "QEP-0-QGIS-data-provider-for-Dameng.md". We will also include details on the technical ramifications brought by this proposal, such as extra dependencies to QGIS and potential tradeoffs for using Dameng. 

In the meantime, we are very interested in the Python plugin approach. May we ask if it would be possible to consider acceptance and integration through a Python plugin?

Once again, thank you for your invaluable feedback and suggestions. We look forward to an extended collaboration with the QGIS community to contribute to the development of QGIS.

troopa81:
> In the meantime, we are very interested in the Python plugin approach. May we ask if it would be possible to consider acceptance and integration through a Python plugin?

Any python plugins can be [uploaded](https://plugins.qgis.org/plugins/add/) in the [plugin platform](https://plugins.qgis.org/)  and could be then installed by any user using the extension widget.

PR #328, opened 2025-02-13
QEP 322: Point layer edition from the profile tool

ptitjano:
Here is a screenshot of what the solution could look like: 

Capture d’écran du 2025-02-13 16-18-46

### DIFF
--- a/qep-322-profile-tool-point-edition.md
+++ b/qep-322-profile-tool-point-edition.md
@@ -1,0 +1,60 @@
+# QGIS Enhancement 322: Add support for point layer edition from the profile tool
+
+**Date** 2025/02/13
+
+**Author** Jean Felder (@ptitjano), Jacky Volpes (@Djedouas)
+
+**Contact** jean dot felder at oslandia dot com, jacky dot volpes at oslandia dot com
+
+**Version** QGIS 3.44
+
+# Summary
+
+This proposal aims to enable the editing of point layers directly from the elevation profile. It will be possible to add new points, move existing ones, and delete them. Additionally, selecting points from the elevation profile will be possible. Editing directly within the elevation profile will make it easier to precisely adjust the elevation of 3D points.
+
+## Proposed Solution
+
+The main challenge of this feature is ensuring that these new functionalities do not conflict with the existing editing options in the main canvas. To make the solution as clear as possible, we propose to:
+
+- Add a new editing section to the toolbar within the elevation profile panel.
+- Display the editing icon for the layer currently selected in the elevation profile tree view.
+- Synchronize edition state between the main canvas and the elevation profile: enabling edition on a layer from the elevation profile will activate it in the main canvas, and vice versa.
+
+Tolerance will not be handled when adding or moving points. This means that the points will always be placed on the profile line.  
+This editing toolbox will not be enabled if the selected layer is not a point layer.
+
+The different operations (Add, Delete, Move, Select) will be implemented by creating new tools that inherit from `QgsPlotTool`.
+
+This branch already contains a rough implementation: https://github.com/ptitjano/QGIS/tree/capture-profile-points
+
+## Deliverables
+
+- A new editing toolbox in the profile tool
+- Add / Move / Delete / Select points from the profile tool
+
+### Affected Files
+
+- src/app/elevation/qgselevationprofilewidget.{h,cpp}
+- src/app/elevation/qgselevationprofiletooladdpoint.{h,cpp} (new files)
+- src/app/elevation/qgselevationprofiletoolmovepoint.{h,cpp} (new files)
+- src/app/elevation/qgselevationprofiletoolselectfeatures.{h,cpp} (new files)
+- src/core/elevation/qgsprofilerenderer.cpp
+- src/core/layout/qgslayoutitemelevationprofile.cpp
+- src/gui/elevation/qgselevationprofilelayertreeview.cpp
+- src/gui/elevation/qgselevationprofilecanvas.cpp
+- src/gui/plot/qgsplotrubberband.{h,cpp} (new files)
+
+## Risks
+
+Increased complexity and maintenance burden in the profile tool code.
+
+## Performance Implications
+
+N/A
+
+## Backwards Compatibility
+
+Not Applicable.
+
+
+Funded by: BRGM


qep-322-profile-tool-point-edition.md
```diff
@@ -0,0 +1,60 @@
...
+
+**Author** Jean Felder (@ptitjano), Jacky Volpes (@Djedouas)
+
+**Contact** jean dot felder at oslandia dot com, jacky dot volpes at oslandia dot com
+
+**Version** QGIS 3.44
+
+# Summary
+
+This proposal aims to enable the editing of point layers directly from the elevation profile. It will be possible to add new points, move existing ones, and delete them. Additionally, selecting points from the elevation profile will be possible. Editing directly within the elevation profile will make it easier to precisely adjust the elevation of 3D points.
```
nyalldawson, line 13: How would deletion work? Would it work on selection, or be a "click on points to delete" tool? If the latter, is it click per individual point? or marquee based?

qep-322-profile-tool-point-edition.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Proposed Solution
+
+The main challenge of this feature is ensuring that these new functionalities do not conflict with the existing editing options in the main canvas. To make the solution as clear as possible, we propose to:
+
+- Add a new editing section to the toolbar within the elevation profile panel.
```
nyalldawson, line 19: For the same reasons as the pushback against the editing tools in the 3d window toolbar, can we avoid just dumping all these actions into the existing profile window toolbar? We need an alternative approach which doesn't add extra UI complexity (and minimum horizontal window size requirement).

qep-322-profile-tool-point-edition.md
```diff
@@ -0,0 +1,60 @@
...
+- Display the editing icon for the layer currently selected in the elevation profile tree view.
+- Synchronize edition state between the main canvas and the elevation profile: enabling edition on a layer from the elevation profile will activate it in the main canvas, and vice versa.
+
+Tolerance will not be handled when adding or moving points. This means that the points will always be placed on the profile line.
```
nyalldawson, line 23: Can you clarify what you mean here? Do you tolerance when picking the point to move? Or tolerance in allowing a user to "tear away" a point from the profile?

If the latter, wouldn't that be a handy feature? The point would keep the same x/y location, but could be translated up or down in the z axis...

qep-322-profile-tool-point-edition.md
nyalldawson, line 13: Presumably selection is the same between elevation plots + the main layer selection? (Can you explicitly state this)

uclaros:
If you just want to snap points to the profile line, wouldn't it be simpler to just use the _Drape (set Z value from raster)_ algorithm in _Edit features in place_ mode?

qep-322-profile-tool-point-edition.md
ptitjano: I agree that this is probably the most controversial part of this proposal. 

However, we did some user testing and this is only approach that made sense for them. Indeed, when using this feature, they need to open to open several elevation profile widget. From their point of view, each widget is a standalone application. Thus, it needs their own buttons.

qep-322-profile-tool-point-edition.md
ptitjano: It will be selection based. I have added some text to explain it.

qep-322-profile-tool-point-edition.md
ptitjano: I have tried to explain it in this new version. It will be possible to ` keep the same x/y location, but could be translated up or down in the z axis`

ptitjano:
> I'm guessing the attribute form will immediately open after adding points, just like it does for 2d maps?

Yes. Added some text to explain it.

ptitjano:
> If you just want to snap points to the profile line, wouldn't it be simpler to just use the _Drape (set Z value from raster)_ algorithm in _Edit features in place_ mode?

Indeed. This is just a different workflow than what is proposed here.

wonder-sk:
This is an interesting one - when building point cloud editing tools, we were considering whether to have editing tools available in the elevation profile views as well. In the end we came to the conclusion that elevation profiles are probably best left as pure read-only views of data, and for any editing we should stick to 2d/3d canvas - for several reasons:
- duplication of efforts / functionality - the same editing tools being implemented multiple times, with different quirks
- 3d views offer superset of functionality compared to profile views (okay, we still need to sort out easy setup of clipping box and fix orthogonal camera view, but we're nearly there)
- many kinds of edits are simply not possible / not well-defined in profile view. For example, line or polygon editing probably won't make any sense in profiles. Or the fact that you suggest that points would get snapped to the profile line may be surprising for users if they only want to change z-value of a point

How do you plan to support snapping - would it respect snapping as set up for the 2d canvas, or would there be some special rules?

How about multiple selection - would it be possible to select multiple points and move them at once?

Djedouas:
> If you just want to snap points to the profile line, wouldn't it be simpler to just use the _Drape (set Z value from raster)_ algorithm in _Edit features in place_ mode?

This proposal aims to give more flexibility to the user. It will be possible to fine tune the Z coordinate per point, snap it on a DEM or completely another layer. It will also be possible to move (slide) a point along the profile. And more importantly, create new points.

qep-322-profile-tool-point-edition.md
nyalldawson: > However, we did some user testing and this is only approach that made sense for them. Indeed, when using this feature, they need to open to open several elevation profile widget. From their point of view, each widget is a standalone application. Thus, it needs their own buttons.

We still need some alternative vs just chucking them in the toolbar. It could be an additional (hidden by default) toolbar in the profile window, or a drop-down menu with edit actions, or some initially invisible panel. But I'm -1 to what you've described here.

qep-322-profile-tool-point-edition.md
```diff
@@ -0,0 +1,84 @@
...
+
+In editing mode, activate it by clicking the "Add Point Feature" button in the elevation profile panel. then click on the elevation profile to add a new point. If necessary, the attribute form will immediately open. Snapping will be handled.
+
+### Move features
+
+In editing mode, activate it by clicking the "Move Points Feature" button in the elevation profile panel. Select a point by clicking on it and then click on its new location. Moving points will have a special "lock abscissa" keyboard modifier to avoid changing x/y coordinate while moving. It will also handle snapping. 
+
+### Delete features
+
+Deletion of features will work on selection. First, select features by clicking on them or using a rectangle selection. Then delete them by clicking the "Delete Features" button from the elevation profile toolbar.
```
nyalldawson, line 45: Personally, I don't see any need for a special delete features tool in the elevation profile -- there's barely any difference vs doing the selection in the profile and then hitting delete on the main window toolbar.

Jakob-Lanstorp: I think the elevation profile should have Add point, Move point and Delete point. This might seem redundant to the main window, but nevertheless, this is a common design also between the the main window and a the Attribute window with several common buttons.

ptitjano:
> This is an interesting one - when building point cloud editing tools, we were considering whether to have editing tools available in the elevation profile views as well. In the end we came to the conclusion that elevation profiles are probably best left as pure read-only views of data, and for any editing we should stick to 2d/3d canvas - for several reasons:
> 
>     * duplication of efforts / functionality - the same editing tools being implemented multiple times, with different quirks
> 
>     * 3d views offer superset of functionality compared to profile views (okay, we still need to sort out easy setup of clipping box and fix orthogonal camera view, but we're nearly there)
> 

Please see the comment the comment from @Jakob-Lanstorp https://github.com/qgis/QGIS-Enhancement-Proposals/pull/328#issuecomment-2659296020  
It explains why this is is a useful addition to the profile tool and the workflow.

>     * many kinds of edits are simply not possible / not well-defined in profile view. For example, line or polygon editing probably won't make any sense in profiles. Or the fact that you suggest that points would get snapped to the profile line may be surprising for users if they only want to change z-value of a point

This is why those tools would only be available for points. As you explain, this does not make sense for lines or polygons.
> 
> 
> How do you plan to support snapping - would it respect snapping as set up for the 2d canvas, or would there be some special rules?

The profile tool already has its own snapping. This would use it.

> 
> How about multiple selection - would it be possible to select multiple points and move them at once?

Moving multiple points at once is not planned. I guess it could be added later on.

Jakob-Lanstorp:
Agree only points should be editable in elevation profile, not line or polygon geometry. Also only one point, not multiple points, to be move at the same time, but multiple point should be able to be delete at once if they are selected. I think snapping could be ignored in this iteration. What I mean by snapping is that the new point added in the elevation profile, gets the real x,y coordinated (from an existing geometry), instead of being projected on the profile line x,y. If you have a large profile tolerance on the profile, then there might be a large distance from point to the perpendicular projection to the profile line. This could be handled by the user instead if they make sure that the profile goes directly through the "snapped" points. If this is not clear I will be happy to eleborate more on the subject.

qep-322-profile-tool-point-edition.md
T4mmi: +1 for the "per-profile" toolbar ...  profile "layer tree view" is NOT synced with the main one !

The main problem happens when you select the layer you wanna edit in the profile window, and the "toogle editing" button of the main window ... toogles the main window current layer (which is NOT the one you selected in your profile). This is really enoying at use, since whenever you want to edit something in your profile you have to go back to the main window (exit dual screen ergonomy).

The "feel good" behavior is *edit what you select ... in your current/active widget*

nyalldawson: @T4mmi that's not my issue though -- my issue is with adding 7 more icons to the elevation profile toolbar, and the associated UI complexity/bloat. I think at a minimum these should be hidden by default, and only shown if the user opts-in via eg a "Show Editing Tools" toggle in the profile settings drop-down menu.

Jakob-Lanstorp: Hiding icons as default is fine when the setting can be toggled.

nicogodet:
+1 too 

We regularly deals with survey data (terrestrial or bathymetric data) and having the possibility to edit/create them in elevation profile would be a nice feature.
The current use case I see is to create a point layer which represents the top of a dike/high ground from point cloud layer (human eye is better than machine to recreate correct elevation from a noisy point cloud).
Workflow would be:
- create a line or directly draw profile line where the crest line should be
- create/edit point layer directly in profile elevation view and point are created along profile line.

France is currently covered with more than 1 000 000 000 000 points in open data, this feature would open a new door on how we could exploit all these informations.

Jakob-Lanstorp:
> > There is a tight relation between the point to edit in the profile window and the profile line.
> > Depending on the profile line buffer i.e. the profile Tolerance setting, a point within the tolerance is projected to the profile line to appear on the profile.
> > So when moving an existing point in the profile window, it would be "reprojected" to the profile line. If the tolerance is 100m, then by moving a point 1m in the profile window might actually move it 100m in the 2D-Window, based of the tolerance.
> > In geological software I solve this by having the profile line go trough the point to edit. This might not be easy with a point cloud. Anyways it should be clear what happens.
> 
> +1 for the "it should be clear what happens ! I think there are two options :
> 
> * editing in profile view ONLY modify the $z component (no reprojection)
> * editing a projected point will snap it to the profile (modify all $geometry.coords)

Both options would be nice to have.

ptitjano:
> If I try to summarize both options, it sums up to either :
> 
>     1. Add the displacement vector(s) in the profile to the original point(s)
> 
>     2. Use blunt final coordinates as replacements for the original point(s)
> 
> 
> after reflexions, forcing the reprojection inside the profile (2) simplifies things since it handles moves in both profiles directions ! option (1) needs to match points before/after, plus it would induce some headaches around the pivot points (in case the profile is based on polylines...)

The QEP has been updated to clarify that the points will be reprojected. This does not preclude having both options available in a future development.

ptitjano:
From my perspective, all previously raised questions and comments have been addressed. Are there any additional questions to discuss?

benoitdm-oslandia:
@nyalldawson it seems we reach the quorum. Are you ok to merge?

PR #330, opened 2025-02-23
QEP 314: Relax prohibition against "auto" - exception for use of auto for variables initialized using casts

nyalldawson:
This change relaxes the prohibition against auto to allow use of auto for variable types when the variable type is explicitly stated during its initialization as the result of a cast operation, such as ``dynamic_cast``, ``static_cast``, ``const_cast``, ``reinterpret_cast``, ``qobject_cast``,
    ``qgsgeometry_cast``, or ``qgis::down_cast``

Specifically:

    // allowed, the pointer types are explicit during initialization:
    auto markerSymbol = dynamic_cast< QgsMarkerSymbol* >( symbol );
    auto vectorLayer = qobject_cast< QgsVectorLayer* >( layer );
    if ( auto lineString = qgsgeometry_cast< const QgsLineString* >( geometry ) ) { ... }

    // NOT allowed, the type is not explicit:
    auto myLineStringPtr = downcastGeometryAsLineString( geometry );

Refs comment at https://github.com/qgis/QGIS-Enhancement-Proposals/pull/319#issuecomment-2652037279

### DIFF
--- a/qep-314-coding-style.md
+++ b/qep-314-coding-style.md
@@ -88,6 +88,19 @@ Notes:
   // NOT allowed, the unique_ptr type is not explicit:
   auto myUniquePtr = createObjectFunction( 0, 5 );
   ```
+  - 3.1.3. ``auto`` may be used for variable types where the variable type is explicitly stated via a cast operation, including
+    ``dynamic_cast``, ``static_cast``, ``const_cast``, ``reinterpret_cast``, ``qobject_cast``,
+    ``qgsgeometry_cast``, or ``qgis::down_cast``. Eg
+
+  ``` cpp
+  // allowed, the pointer types are explicit during initialization:
+  auto markerSymbol = dynamic_cast< QgsMarkerSymbol* >( symbol );
+  auto vectorLayer = qobject_cast< QgsVectorLayer* >( layer );
+  if ( auto lineString = qgsgeometry_cast< const QgsLineString* >( geometry ) ) { ... }
+
+  // NOT allowed, the type is not explicit:
+  auto myLineStringPtr = downcastGeometryAsLineString( geometry );
+  ```
 
 - 3.2. If ``enums`` are to be used outside of a single .h/.cpp file (or there is a reasonable chance that they will be in future!), they should be placed inside the ``Qgis`` namespace.
 


m-kuhn:
Comparing the examples

case 1

```cpp
auto markerSymbol = dynamic_cast< QgsMarkerSymbol* >( symbol );
```
vs.
```cpp
QgsMarkerSymbol* markerSymbol = dynamic_cast< QgsMarkerSymbol* >( symbol );
```

case 2

```cpp
auto vectorLayer = qobject_cast< QgsVectorLayer* >( layer );
```
vs.
```cpp
QgsVectorLayer* vectorLayer = qobject_cast< QgsVectorLayer* >( layer );
```

case 3

```cpp
if ( auto lineString = qgsgeometry_cast< const QgsLineString* >( geometry ) ) { ... }
```

vs.

```cpp
if ( const QgsLineString* lineString = qgsgeometry_cast< const QgsLineString* >( geometry ) ) { ... }
```

I personally always prefer the explicit version over the auto version, it provides me with more information compared to the "auto" version and I also don't see auto remove a lot of clutter here.

uclaros:
The class names used in the examples are small enough to not make a big difference between explicit and auto.
When using longer classes though, the difference is also bigger:
```cpp
const QgsVectorLayerUndoPassthroughCommandChangeGeometry *merge = dynamic_cast<const QgsVectorLayerUndoPassthroughCommandChangeGeometry *>( other );
```
vs
```cpp
auto merge = dynamic_cast<const QgsVectorLayerUndoPassthroughCommandChangeGeometry *>( other );
```

uclaros:
Also wanted to add that to me it makes little sense to allow `auto` for `std::make_unique` and not for casts, as syntactically the two cases are the same (the casts case is actually slightly cleaner, since the type within the angle brackets is copied verbatim).

nyalldawson:
This is ready for voting now

uclaros:
+1

m-kuhn:
-0

rouault:
+1

alexbruy:
+1

nyalldawson:
Declaring as passed with approval from @alexbruy , @rouault , @elpaso, @uclaros

PR #338, opened 2025-03-25
QEP 338: SIP Incremental builds

dvdkon:
Currently QGIS uses SIP to generate its Python bindings, doing a full rebuild every time any relevant header file is changed. This can take over ten minutes, making work on the header files painfully slow. This is exacerbated by a regression in SIP's performance since SIP 5.

I propose adding the option to generate one binding file per source header file, meaning that only a small part of the bindings will need to be rebuilt on each change.  
More technical details are in the QEP proper.

NOTE: This QEP is part of a grant proposal for the 2025 round. Could someone please tag it?


nyalldawson:
@dvdkon can you please announce this to the qgis-dev and qgis-psc lists?

QEP-338-SIP-sip-incremental-builds.md
```diff
@@ -0,0 +1,64 @@
...
+
+I propose adding the option of building Python bindings incrementally -- per
+each header. This should speed up the Python bindings build during development
+by orders of magnitude.
+
+## Proposed Solution
+
+SIP 6's `sipbuild` would be used as a Python module in a script that generates
+one C++ source file per header file. Currently, `sip-build` generates one file
+per class, optionally concatenating them into some fixed number of files, but
```
troopa81, line 32: the concatenate option is broken, and produce some time, non compiling code, depending on where it split the code. If you manage to have a C++ file per header file, I think we should completely get rid of this option, and have instead one cpp file per module, or one cpp file per header file.

QEP-338-SIP-sip-incremental-builds.md
```diff
@@ -0,0 +1,64 @@
...
+the SIP Python API is considerably more flexible.
+
+A new CMake variable would be introduced that switches the bindings build
```
troopa81, line 35: Why not link it to the [ENABLE_UNITY_BUILDS](https://github.com/qgis/QGIS/blob/201e6bae76c1c967ca617864dd854920f34eecdc/CMakeLists.txt#L1121) option ?

troopa81 (commented):
My main concern is whether it is easily doable or not. But if it is, I think it would be a very nice addition.

Did you consider contributing to SIP project add adding an option "one source per sip file" ?

This way, we wouldn't no suffer possible Python API break.

QEP-338-SIP-sip-incremental-builds.md
dvdkon: I managed to fix at least one problem with the concatenate option in a [previous PR](https://github.com/qgis/QGIS/pull/59939#issuecomment-2569110596). The issue was that QGIS' .sip files contained code that didn't compile on its own, so the concatenation really masked the issue more than broke anything.  
I haven't heard of any problems since then, but maybe I just didn't notice.

Generating more files will surely slow down clean builds, but if it turns out to be negligible I wouldn't mind keeping only the new build option.

QEP-338-SIP-sip-incremental-builds.md
dvdkon: I don't think having options that control multiple somewhat-related things is a good idea. I'd certainly want to e.g. test my SIP changes without also turning on unity builds.

Furthermore, as a user I'd expect `ENABLE_UNITY_BUILDS` to enable unity builds and do nothing else.

dvdkon:
> Did you consider contributing to SIP project add adding an option "one source per sip file" ?

Right now, SIP only accepts one `.sip` file per module, so QGIS generates a file that `%Includes` every other file. Untangling this will require some work on the QGIS side, as will adapting the CMake scripts. But if there's interest in merging this option into SIP directly, I'm not opposed.  
The current proposal mostly follows the "path of least resistance", but that's certainly not necessary.

nyalldawson:
I'm a big +1 to this! (Even if it ultimately proves to be infeasible and a dead-end.).

The current speed is just far too prohibitive to efficient development, and any chance of improving this is going to give huge rewards.

QEP-338-SIP-sip-incremental-builds.md
troopa81: > I haven't heard of any problems since then, but maybe I just didn't notice.

Just try to reduce the SIP_CONCAT_PARTS number and you'll see :) We are increasing this number only because it breaks build when the API is modified

> Generating more files will surely slow down clean builds, but if it turns out to be negligible I wouldn't mind keeping only the new build option.

If we have incremental build with one file per header file for development and a "unity" one for CI/packaging, I think we are good and don't really see what would be the advantage of the concatened one

QEP-338-SIP-sip-incremental-builds.md
troopa81: OK. It looks kind of the same thing to me, either a per header file build or a big-all-file-in-one-cpp unity build.

troopa81:
> Right now, SIP only accepts one .sip file per module, so QGIS generates a file that %Includes every other file. Untangling this will require some work on the QGIS side, as will adapting the CMake scripts. But if there's interest in merging this option into SIP directly, I'm not opposed.
The current proposal mostly follows the "path of least resistance", but that's certainly not necessary.

Yes, you're right!

anitagraser:
Thank you for submitting your proposal to the [2025 QGIS Grant Programme](https://github.com/qgis/PSC/issues/63). The 2 week discussion period ends today. At the end of the discussion, the proposal author has to provide a 3-line pitch of their proposal for the voter information material. (For an example from last year check https://github.com/qgis/PSC/issues/58#issuecomment-1567892412)

dvdkon:
@anitagraser Thanks, here's my pitch:

Currently, building the Python bindings is a slow process, since for any small change in a relevant header it rebuilds the whole bindings. This proposal suggests adding an option to build the bindings incrementally, per-file, by using SIP's Python API. This should speed up some incremental builds by orders of magnitude.

nyalldawson:
To kick off the formal technical votes I'll add my +1

3nids:
+1

NathanW2:
+1

QEP-338-SIP-sip-incremental-builds.md
m-kuhn: Isn't this technically very, very similar to a unity build?

alexbruy:
+1

QEP-338-SIP-sip-incremental-builds.md
```diff
@@ -0,0 +1,64 @@
...
+process from the current `sip-build`-based one to using the new script.
+
+## Deliverables
+
+- Python script that generates Python bindings per-file, rather than
+  per-project.
```
m-kuhn, line 41: Would this python script be released on pypi to potentially attract more users or will it be very QGIS specific?

troopa81:
+1

elpaso:
+1

ptitjano:
+1

Samweli:
+1

nyalldawson:
Declaring as passed with approval from @troopa81 , @elpaso , @ptitjano , @nyalldawson , @NathanW2 @3nids , @alexbruy

nirvn:
Glad this was accepted, improvements on this front can't come soon enough. As part of the bug fixing round for QGIS 3.44, I've spent a day or two working against a local Qt6 build (which forces us to move to SIP 6), and the time to rebuild simple changes makes it nearly impossible to working with when hoping from one branch to another.

dvdkon:
> Dear David, this is a friendly reminder that the deadline for follow-up reports is approaching (2025-10-30). Please send the report to the PSC mailing list and the invoice to finance / Andreas so we can process the payments.

@anitagraser Hi, I've already sent the [report](https://lists.osgeo.org/pipermail/qgis-psc/2025-October/010660.html) to the mailing list, hopefully that is enough. The invoice should be handled by Lutra.

(And sorry for the late reply, apparently my last comment didn't go through?)

PR #339, opened 2025-04-30
QEP 339: Processing Algorithm to Convert ASCII Point Cloud Files

ptitjano:
This is a proposal to add a processing algorithm to convert ASCII Point Cloud Files  

This is based on the feedback from https://github.com/qgis/QGIS/pull/61498


wonder-sk:
I may be missing something, but why would we want to write another text file? If the aim is to get the original file loaded in QGIS, then why not convert it from text to COPC (using `writers.copc`), so that we can skip another conversion pass with `untwine`?

PR #341, opened 2025-06-18
Add QEP 341: add 3D pedestrian view

benoitdm-oslandia:
Here is an screencast (without the cursor capture) of a POC:

[3D pedestrian view screencast](https://github.com/user-attachments/assets/9f839c8f-fda4-45f1-93fd-a54bfe20b454)

cc @ptitjano

### DIFF
--- a/qep-341-3d-pedestrian-view.md
+++ b/qep-341-3d-pedestrian-view.md
@@ -18,17 +18,17 @@ This mode can be activated by selecting a specific point or by default from the 
 
 ### Enabling and disabling this tool
 
-A button will be added to the 3D view toolbar to activate this tool and select where to begin navigation. Once the button is clicked (it stays pressed), a marker will be placed vertically above the center of the 3D view. The user moves the marker to the desired location and confirms with a left mouse click or the Enter key.
+A drop-down menu (with the 3 navigation modes) in the "camera" tool in the toolbar to easy the switching from one mode to another.
 
-Once the mode is activated:
+Once this mode is activated, the user will select where to begin navigation: a marker will be placed vertically above the center of the 3D view. The user moves the marker to the desired location and confirms with a left mouse click or the Enter key.
+
+Once the user click to select the starting point:
 
 * the camera moves to a human-like height above the terrain at the selected location
 * the camera is facing the same direction as before activation
 * the movement mode changes (see below)
 
-The user exits this mode by clicking the previous toolbar button (still pressed) or by pressing the *Esc* key or by doing a right click with the mouse.
-
-Optionally, the button can be placed in the Camera menu but since no pop-up or dialog box is displayed, the user will lose the indication that they have switched to that tool and will not know how to exit that tool.
+The user exits this mode by switching to another navigation mode or by pressing the *Esc* key or by doing a right click with the mouse.
 
 ### Movements
 
@@ -55,6 +55,10 @@ Here is the whole mouse mapping:
 The current **Walk** mode will be renamed **Fly** as the pedestrian view is more a real walk mode.
 
 As we will have 3 navigation modes, the 3D UI should display the current mode in use and a way to switch from one to another.
+
+### Integration
+
+The pedestrian view map tool will handle the location picking stage (with a rubberband) in its dedicated class file (something like `QgsWalkModeMapTool`) and this map tool will call API functions at the "app" level to activate the walk mode at the user specified location (and set it back at the previous mode when it quits).
 
 ## Deliverables
 
--- a/qep-341-3d-pedestrian-view.md
+++ b/qep-341-3d-pedestrian-view.md
@@ -1,0 +1,78 @@
+# QGIS Enhancement: Pedestrian walk mode in the 3D Map View
+
+**Date** 2025/06/18
+
+**Author** Benoit De Mezzo ([@benoitdm-oslandia](https://github.com/benoitdm-oslandia)), Jean Felder ([@ptitjano](https://github.com/ptitjano))
+
+**Contact** benoit dot de dot mezzo at oslandia dot com, jean dot felder at oslandia dot com
+
+**Version** QGIS 4.X
+
+## Summary
+
+This QGIS enhancement adds a human-level navigation mode to the 3D view (like Google Street View). It complements the two existing navigation modes.
+
+This mode can be activated by selecting a specific point or by default from the center of the map. Once this navigation mode is activated, the camera's elevation is locked at 1.80 meters above the ground. This height can be configured in the 3D view configuration dialog.
+
+When this mode is activated, a small inset with the 2D view is superimposed on the 3D view.
+
+## Proposed Solution
+
+### Enabling and disabling this tool
+
+A button will be added to the 3D view toolbar to activate this tool and select where to begin navigation. Once the button is clicked (it stays pressed), a marker will be placed vertically above the center of the 3D view. The user moves the marker to the desired location and confirms with a left mouse click or the Enter key.
+
+Once the mode is activated:
+
+* the camera moves to a human-like height above the terrain at the selected location
+* the camera is facing the same direction as before activation
+* the 2D map inset is displayed (see below)
+* the movement mode changes (see below)
+
+The user exits this mode by clicking the previous toolbar button (still pressed) or by pressing the *Esc* key or by doing a right click with the mouse.
+
+Optionally, the button can be placed in the Camera menu but since no pop-up or dialog box is displayed, the user will lose the indication that they have switched to that tool and will not know how to exit that tool.
+
+### Movements
+
+Movements can be done using the keyboard or mouse and are free: unconstrained by obstacles or roads.
+
+Forward and backward movement will be controlled by the up and down keys (and mouse wheel). Lateral movement will be controlled by the left and right keys (and lateral movement of the mouse wheel). Lateral mouse movements will allow you to turn without moving. Vertical mouse movements allow you to look up or down.
+
+### 2D view inset
+
+When this mode is enabled, an inset with the 2D view is displayed as an overlay to help the user navigation. It is centered on the user's current X/Y position, and its extent is adjusted based on the camera's tilt (the larger the far plane, the larger the extent).
+
+This inset could also be a standalone feature enabled from the 3D view settings ("Navigation Synchronization").
+
+As in Google Street View, the position and size of this inset would be fixed.
+
+## Deliverables
+
+* 2D view inset class
+* new maptool class
+* new toolbar button image
+
+### Example(s)
+
+See associated PR.
+
+### Affected Files
+
+*(optional)*
+
+## Risks
+
+None
+
+## Performance Implications
+
+None
+
+## Further Considerations/Improvements
+
+* identification
+
+## Backwards Compatibility
+
+Yes
--- a/qep-341-3d-pedestrian-view.md
+++ b/qep-341-3d-pedestrian-view.md
@@ -24,7 +24,6 @@ Once the mode is activated:
 
 * the camera moves to a human-like height above the terrain at the selected location
 * the camera is facing the same direction as before activation
-* the 2D map inset is displayed (see below)
 * the movement mode changes (see below)
 
 The user exits this mode by clicking the previous toolbar button (still pressed) or by pressing the *Esc* key or by doing a right click with the mouse.
--- a/qep-341-3d-pedestrian-view.md
+++ b/qep-341-3d-pedestrian-view.md
@@ -39,6 +39,10 @@ Movements can be done using the keyboard or mouse and are free: unconstrained by
 
 Forward and backward movement will be controlled by the up and down keys (and mouse wheel). Lateral movement will be controlled by the left and right keys (and lateral movement of the mouse wheel). Lateral mouse movements will allow you to turn without moving. Vertical mouse movements allow you to look up or down.
 
+In order to match existing, binding to W/A/S/D keys (for forward, left, backward and right) will also be added.
+
+Movement speed can be changed by holding `Ctrl` (decrease speed) or `Shift` (increase speed) keys.
+
 ### 2D view inset
 
 When this mode is enabled, an inset with the 2D view is displayed as an overlay to help the user navigation. It is centered on the user's current X/Y position, and its extent is adjusted based on the camera's tilt (the larger the far plane, the larger the extent).
@@ -46,6 +50,14 @@ When this mode is enabled, an inset with the 2D view is displayed as an overlay 
 This inset could also be a standalone feature enabled from the 3D view settings ("Navigation Synchronization").
 
 As in Google Street View, the position and size of this inset would be fixed.
+
+See <https://github.com/qgis/QGIS-Enhancement-Proposals/pull/341#discussion_r2156539364>.
+
+### Existing navigation modes
+
+The current **Walk** mode should be renamed **Fly** as the pedestrian view is more a real walk mode.
+
+As we will have 3 navigation modes the 3D UI should display the current mode in use and a way to switch from one to another.
 
 ## Deliverables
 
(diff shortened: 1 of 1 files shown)


qep-341-3d-pedestrian-view.md
```diff
@@ -0,0 +1,78 @@
...
+
+The user exits this mode by clicking the previous toolbar button (still pressed) or by pressing the *Esc* key or by doing a right click with the mouse.
+
+Optionally, the button can be placed in the Camera menu but since no pop-up or dialog box is displayed, the user will lose the indication that they have switched to that tool and will not know how to exit that tool.
+
+### Movements
+
+Movements can be done using the keyboard or mouse and are free: unconstrained by obstacles or roads.
+
+Forward and backward movement will be controlled by the up and down keys (and mouse wheel). Lateral movement will be controlled by the left and right keys (and lateral movement of the mouse wheel). Lateral mouse movements will allow you to turn without moving. Vertical mouse movements allow you to look up or down.
```
nyalldawson, line 40: I think:

- We should match this to the existing walk/fly mode, and also accept W/A/S/D keys for movement.
- I don't like use of mouse wheel for forward/backward movement, I don't think that's natural or useful here. I'd rather this were left to control the movement speed to match the existing walk/fly mode. 
- We should match the existing mode and allow ` to toggle between cursor captured mode and left-click-drag modes.
- Also need to make sure that the invert vertical axis setting is respected for this mode too.

qep-341-3d-pedestrian-view.md
```diff
@@ -0,0 +1,78 @@
...
+
+### 2D view inset
+
+When this mode is enabled, an inset with the 2D view is displayed as an overlay to help the user navigation. It is centered on the user's current X/Y position, and its extent is adjusted based on the camera's tilt (the larger the far plane, the larger the extent).
+
+This inset could also be a standalone feature enabled from the 3D view settings ("Navigation Synchronization").
+
+As in Google Street View, the position and size of this inset would be fixed.
```
nyalldawson, line 48: What's the benefit in this over just using the existing option to show 3d map view extent in the 2d canvas? That's already there, so dropping this would reduce complexity considerably. I also think this forced overlay would be annoying in some use cases where you want walk mode but don't want any part of the view obscured.

qep-341-3d-pedestrian-view.md
```diff
@@ -0,0 +1,78 @@
...
+
+None
+
+## Further Considerations/Improvements
+
+* identification
+
+## Backwards Compatibility
+
+Yes
```
nyalldawson, line 78: Where are the technical changes described? Or was this opened early to get feedback on the UX and you'll populate those for discussion later?

qep-341-3d-pedestrian-view.md
```diff
@@ -0,0 +1,78 @@
...
+## Summary
+
+This QGIS enhancement adds a human-level navigation mode to the 3D view (like Google Street View). It complements the two existing navigation modes.
+
+This mode can be activated by selecting a specific point or by default from the center of the map. Once this navigation mode is activated, the camera's elevation is locked at 1.80 meters above the ground. This height can be configured in the 3D view configuration dialog.
+
+When this mode is activated, a small inset with the 2D view is superimposed on the 3D view.
+
+## Proposed Solution
+
```
nyalldawson, line 20: Can you add some notes about how this relates to the existing "Walk" mode?

In my view we should rename the existing mode to "Fly" or "Hover", in both the UI and API, and then use "Walk" to describe the new mode. Functionally it should operate the same, but with the additional constraint of the fixed height above terrain.

nyalldawson:
@benoitdm-oslandia please announce this to the DEV and USER lists, as per https://github.com/qgis/QGIS-Enhancement-Proposals?tab=readme-ov-file#process-and-policies

qep-341-3d-pedestrian-view.md
benoitdm-oslandia: We already made a POC and the changes have low/no impact on the existing code. We will populate the QEP when more feedback are received.

qep-341-3d-pedestrian-view.md
kannes: A big +1 for WASD from me. It is a well established schema for avatar navigation, not just in gaming. E.g. Google Earth and Blender support it too.

For azerty support it should be enough if users can customize the key bindings to ZQSD.

qep-341-3d-pedestrian-view.md
```diff
@@ -0,0 +1,90 @@
...
+### Movements
+
+Movements can be done using the keyboard or mouse and are free: unconstrained by obstacles or roads.
+
+Forward and backward movement will be controlled by the up and down keys (and mouse wheel). Lateral movement will be controlled by the left and right keys (and lateral movement of the mouse wheel). Lateral mouse movements will allow you to turn without moving. Vertical mouse movements allow you to look up or down.
+
+In order to match existing, binding to W/A/S/D keys (for forward, left, backward and right) will also be added.
+
+Movement speed can be changed by holding `Ctrl` (decrease speed) or `Shift` (increase speed) keys.
```
kannes, line 44: Just a comment: Nice, this matches common FPS game defaults from my experience.

qep-341-3d-pedestrian-view.md
nirvn: On QWERTY keyboards, +1 for WASD. We should definitively support other keyboard layouts though (AZERTY, DVORAK, etc.), that could be done via a keyboard layout combobox setting, or allow for customization of keyboard shortcuts there.

qep-341-3d-pedestrian-view.md
nyalldawson: Can i suggest that the overlay window is de-coupled from the pedestrian view feature and proposal, and instead implemented as an independent option/proposal? There's evidence here that it's NOT always desirable for pedestrian mode, and it's likely that it IS desirable in other (non-pedestrian) modes too.

qep-341-3d-pedestrian-view.md
nyalldawson: I don't think walk mode should be a map tool. But I see LOTs of overlap between the existing walk (fly) mode and pedestrian mode, as pedestrian is effectively just walk/fly mode with the height constraint on top.

I'm hoping that for the camera controller changes there'll be minimal code introduced by the new mode as a result, and with most of the logic shared with walk/fly mode.

The map tool/mode toggle logic can all sit up at an app level and handle the pedestrian specific map tool logic there.

qep-341-3d-pedestrian-view.md
kannes: Yeah, but I considered Slow and Crawl to be the same here as crawling probably isn't necessary in QGIS. 😄

nyalldawson: Let's just go the whole way and make space jump too 🤣

qep-341-3d-pedestrian-view.md
ptitjano: The map tool here is necessary because one need to click on the 3d view to kickstart the view.  
One could argue that this would be the middle of the view. However, in a lot of cases, this would not work because this point is too far away from the real area of interest. A usual use case of this feature is with a scene which contains roads and building. In that case, this feels natural to click on the road to have a look at the surrounding buildings.

Regarding code duplication between the map tool and the camera controller, one can indeed reuse most of the logic with the existing walk/fly mode. The camera controller would be a consumer of this tool.

One would have the following logic: 
- activate the tool : this displays a marker which allows to precisely select the starting point
- click on a point of the view to kickstart the walk mode
- the events are handled by the QgsCameraController
- exit the view (escape or deactivating the tool): go back to the preview mode and view

wonder-sk:
Technically speaking, what are the differences between the existing walk/fly mode and the proposed pedestrian mode? (i.e. what would change in QgsCameraController?) From what I understand, the main difference is that the new pedestrian mode would try to keep fixed camera elevation above the terrain, while the existing walk/fly mode allows arbitrary elevation? If that's the case, the existing walk/fly mode should just get an optional constraint "keep elevation X meters above terrain", rather than adding a completely new navigation mode. (here I'm talking only about the camera controller code, not GUI.)

How do you plan to keep the camera at constant elevation above the terrain? Are you going to use QgsTerrainGenerator::heightAt() on every move? (note that for raster DEM terrain the implementation is quite naive and using just a very coarse grid)

qep-341-3d-pedestrian-view.md
nyalldawson: > jump, crouch and shoot are probably not needed in a gis 3d window!

Ah yeah, sorry -- I was being tongue in cheek when I mentioned jump. I don't think that's worth the effort!

Instead of crouch/jump I think we could use Q/E as "slide camera vertically", in both fly/pedestrian modes. But I understand if that's out of scope for this work.

qep-341-3d-pedestrian-view.md
benoitdm-oslandia: @nyalldawson what do you think about the comment above?

nyalldawson: I think there was a misunderstanding -- I'm ok with there being a map tool to set the location and switch the mode, but it should be completely disconnected from the actual camera logic for walk mode and the changes in the camera controller. The code for the map tool and navigation mode switch should live up at the "app" level only.

(I.e. we should be able to switch to walk mode via qgis 3d api without being forced to go through a map tool.)

benoitdm-oslandia: so to resume: the pedestrian view map tool will handle the location picking stage (with a rubberband) in its dedicated class file (something like `QgsWalkModeMapTool`) and this map tool will call API functions at the "app" level to activate the walk mode at the specified location (and set it back at the previous mode when it quits).

Can we have something in the UI to improve the user experience? Currently, it is not easy to know which navigation mode we are in or how to switch from one mode to another. Since many users are not very familiar with keyboard shortcuts, it would be nice to add something like a drop-down menu (with the 3 navigation modes) in the "camera" tool in the toolbar. Keyboard shortcuts will still exist for more advanced users.

nyalldawson: > it would be nice to add something like a drop-down menu (with the 3 navigation modes) in the "camera" tool in the toolbar

Totally agree. Exposing all these modes for easy access would be very valuable!

signedav: Just to be sure I understand “app level” correctly. This means that the UI will be implemented in the app (and not in gui or core), which means that it will not be exposed to the QGIS library. Right?

nyalldawson: @signedav that's correct.

benoitdm-oslandia:
@nyalldawson can we call for the vote?

nyalldawson:
@benoitdm-oslandia 
> can we call for the vote?

We can, at any stage you're ready. But in its current form it'd be a -1 from me. I've just added a follow up review with my remaining concerns.

qep-341-3d-pedestrian-view.md
```diff
@@ -0,0 +1,93 @@
...
+* left mouse click: toogle engage/disengage mouse (also bound to QuoteLeft)
+* right mouse click: quit mode (also bound to Esc)
+* Wheel : permanent speed factor (as in current walk mode)
+
+### Existing navigation modes
+
+The current **Walk** mode will be renamed **Fly** as the pedestrian view is more a real walk mode.
+
+As we will have 3 navigation modes, the 3D UI should display the current mode in use and a way to switch from one to another.
+
```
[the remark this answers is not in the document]
uclaros: Also: how will you handle walking into the edges of the terrain/nodata areas?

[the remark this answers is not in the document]
ptitjano: > There's still technical details missing here. I would like to see in depth details on how you propose to handle the fixed height-above-ground requirement. @wonder-sk pointed out earlier that `QgsTerrainGenerator::heightAt()` may be inappropriate due to the coarseness of this API -- have you tested to see whether it works and can give a smooth navigation result?

Indeed. @benoitdm-oslandia has been been working on a prototype to improve `QgsTerrainGenerator::heightAt()` . It mostly involves two changes: 
- listen to the DTM texture generation in `QgsTerrainGenerator` and store them in a cache. The cache has some logic to only keep the highest level when available. This replaces,
- in `Qgs3DMapScene`, when a better level is stored, reload the relevant vector tiles if necessary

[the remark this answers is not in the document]
uclaros: > Not much can be done in that case, except a fallback to `0 + terrainOffset`

Well, you could keep the pedestrian's last elevation so he doesn't fall in potholes/fall off the edge of flat earth :rofl: but rather "float" over them.

Another option would be to interpolate missing values.

qep-341-3d-pedestrian-view.md
```diff
@@ -0,0 +1,93 @@
...
+Movements can be done using the keyboard or mouse and are free: unconstrained by obstacles or roads.
+
+Forward and backward movement will be controlled by the `up` and `down` keys. Lateral movement will be controlled by the `left` and right keys. Lateral mouse movements will allow you to turn without moving. Vertical mouse movements allow you to look up or down.
+
+Here is the whole key mapping:
+
+* ARROW and WASD : forward/backward/shift left/shift right
+  * `Ctrl`: temporary decrease speed
+  * `Shift`: temporary increase speed
+* `Space`: jump **optional**
```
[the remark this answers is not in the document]
benoitdm-oslandia: ok

qep-341-3d-pedestrian-view.md
```diff
@@ -0,0 +1,93 @@
...
+
+Once this mode is activated, the user will select where to begin navigation: a marker will be placed vertically above the center of the 3D view. The user moves the marker to the desired location and confirms with a left mouse click or the Enter key.
+
+Once the user click to select the starting point:
+
+* the camera moves to a human-like height above the terrain at the selected location
+* the camera is facing the same direction as before activation
+* the movement mode changes (see below)
+
+The user exits this mode by switching to another navigation mode or by pressing the *Esc* key or by doing a right click with the mouse.
```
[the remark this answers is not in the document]
benoitdm-oslandia: ok

qep-341-3d-pedestrian-view.md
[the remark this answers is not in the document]
benoitdm-oslandia: We will keep the last elevation to avoid falling in potholes.

qep-341-3d-pedestrian-view.md
```diff
@@ -0,0 +1,93 @@
...
+
+This mode can be activated by selecting a specific point or by default from the center of the map. Once this navigation mode is activated, the camera's elevation is locked at 1.80 meters above the ground. This height can be configured in the 3D view configuration dialog.
+
+## Proposed Solution
+
+### Enabling and disabling this tool
+
+A drop-down menu (with the 3 navigation modes) in the "camera" tool in the toolbar to easy the switching from one mode to another.
+
+Once this mode is activated, the user will select where to begin navigation: a marker will be placed vertically above the center of the 3D view. The user moves the marker to the desired location and confirms with a left mouse click or the Enter key.
```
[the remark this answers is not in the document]
benoitdm-oslandia: > Can you clarify how the marker is moved? Is this via drag and drop?

The marker follows the mouse moves and is placed on the ground where the mouse cursor is located. There is no need to keep the left mouse button to move the marker and the left mouse click validates the starting point. 

> how will you handle the marker and interaction with it?

The marker is a vertical rubberband going upward from the ground with a fixed length (2 points, but it could be a triangle). Each time the user move the mouse, the marker position is updated.

qep-341-3d-pedestrian-view.md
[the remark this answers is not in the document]
nyalldawson: It's ok for me -- but you may want to get those fixes reviewed and merged first, as this proposal depends on them.

Your choice though, we can put this to voting when you're ready.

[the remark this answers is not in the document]
benoitdm-oslandia: the elevation fixes are available in this [PR](https://github.com/qgis/QGIS/pull/62676)

PR #345, opened 2025-07-22
QEP 345: Markers/Hashes line symbol layer blank areas

troopa81:
This QEP proposes to enhance markers and hashes line symbol layer to allow definition of blank areas where no markers or hashes are displayed.

You can read it [here](https://github.com/troopa81/QGIS-Enhancement-Proposals/blob/remove_marker_lines/qep-345-line-symbollayer-blankareas.md)


qep-345-line-symbollayer-blankareas.md
```diff
@@ -0,0 +1,117 @@
...
+
+![issue.png](images/qep345/issue.png)
+
+The rest of the proposal will mention markers but everything describe here applies also to hashes.
+
+## Proposed Solution
+
+In order to have a proper display of the markers along the line, I propose to:
+- Give user a way to identify blank area where markers are not displayed
+- Use the already existing annotation feature to add, move, rotate or even change marker symbology
```
nyalldawson, line 23: There's absolutely no detail about annotations in this QEP. Is this section outdated? Or is this QEP incomplete?

qep-345-line-symbollayer-blankareas.md
```diff
@@ -0,0 +1,117 @@
...
+
+This could look like approximatively this
+
+![poc.gif](images/qep345/poc.gif)
+
+### Fixed marker position
+
+We also propose to add an option "fixed marker position" to fix markers positions whatever the current displayed area and zoom level. Indeed, without this option, you would get this behavior
```
nyalldawson, line 31: I'm -1 to introducing a new option to control this. We already have the user-facing option for "Clip features to Canvas Extent" if this behaviour needs to be user controlled. If not, then I'd rather some heuristics are used to automatically set the the DisableFeatureClipping flag when appropriate (or just always set this for the marker/hash line types). Ideally accompanied with logic to skip the render of the markers/hashes outside the visible extent, to avoid performance issues.

qep-345-line-symbollayer-blankareas.md
```diff
@@ -0,0 +1,117 @@
...
+
+Enabling this option will imply that we disable feature clipping for the given line symbol layer. The flag `Qgis::SymbolLayerFlag::DisableFeatureClipping` would be added to current flags if this option is set.
+
+When set, all the geometry would be drawn entirely and not clipped. This would have performance implication. A tooltip would inform user about this.
+
+### The blank areas
+
+These areas would be defined as start and end distances from the first point of the line or polygon. It would be possible to define the unit of those distances.
+
+A new property would be added on QgsSymbolLayer [ones](https://github.com/qgis/QGIS/blob/d01be3798eb05c74a61629042f7189ad9c594b1e/src/core/symbology/qgssymbollayer.h#L147) and it would be possible to explicity set a data defined property for all `QgsTemplatedLineSymbolLayerBase` symbol layers (which marker line and hash line symbol layer inherits from). It would be possible then to define those blank areas either:
```
nyalldawson, line 45:
```suggestion
A new Property would be added to the QgsSymbolLayer [enum](https://github.com/qgis/QGIS/blob/d01be3798eb05c74a61629042f7189ad9c594b1e/src/core/symbology/qgssymbollayer.h#L147) and it would be possible to explicity set a data defined property for all `QgsTemplatedLineSymbolLayerBase` symbol layers (which marker line and hash line symbol layer inherits from). It would be possible then to define those blank areas either:
```

nyalldawson, line 45: what property type would this be? a string value? If so, how do you propose to store them?  (eg comma separated ranges, ...?)

qep-345-line-symbollayer-blankareas.md
```diff
@@ -0,0 +1,117 @@
...
+- in a static way, 
+- with a specific data field,
+- an auxiliary data storage field,
+- an expression.
```
nyalldawson, line 49: note that if you opt for comma separated strings, then you should also accept expression results which are numeric lists themselves as well.

qep-345-line-symbollayer-blankareas.md
```diff
@@ -0,0 +1,117 @@
...
+
+When used with a field, this information would then be stored per feature, per layer, per symbol layer.
+
+A new widget to edit those blank areas would be inserted within marker line symbol layer widget.
+
+![blankareas_widget.png](images/qep345/blankareas_widget.png)
```
nyalldawson, line 55: This mockup is rather misleading in my opinion -- the list of regions will only be visible and editable IF it's a static list of regions which apply to ALL features. This is a very limited use case! In almost all real-world use cases I'd envisage that per-feature, data-defined regions are in effect... in which case the regions WON'T be visible or editable in this list.

Can you update the proposal to explicitly state this, so that voters can make an informed decision on whether this UX is acceptable?

qep-345-line-symbollayer-blankareas.md
```diff
@@ -0,0 +1,117 @@
...
+
+This option will be false by default which is the actual behavior.
+
+Enabling this option will imply that we disable feature clipping for the given line symbol layer. The flag `Qgis::SymbolLayerFlag::DisableFeatureClipping` would be added to current flags if this option is set.
+
+When set, all the geometry would be drawn entirely and not clipped. This would have performance implication. A tooltip would inform user about this.
+
+### The blank areas
+
+These areas would be defined as start and end distances from the first point of the line or polygon. It would be possible to define the unit of those distances.
```
nyalldawson, line 43: Can you add a note something like "while it will be possible to use screen or page based units like millimeters or inches, it is expected that this will be of limited use as the blank regions would then change size when the map scale is changed. Rather, it is expected users will typically set the region distance unit to map units so that the regions remain fixed in place regardless of the map scale. Accordingly the default unit for the property will be set to Map Units."

qep-345-line-symbollayer-blankareas.md
```diff
@@ -0,0 +1,117 @@
...
+The new map tool will be accessible through the symbology editor (![mActionRemoveAllFromOverview.svg](images/qep345/mActionRemoveAllFromOverview.svg) in the screenshot above, even though a new icon will be proposed), and only here (because you need to choose the symbol layer you are going to add the blank areas).
+
+Button to activate this map tool would be disable if the symbology editor is not docked, because:
+- The properties window is modal and need to be closed before starting to add blank lines
+- User need some direct visual feedback when defining those blank areas
+
+A tooltip will explain why when it's disabled.
+
+This tool will allow to:
+- Click in the neighborhood of the line to start creating an area, click again to finish editing the zone like on the video above.
```
nyalldawson, line 70: which video? did you forget to attach?

qep-345-line-symbollayer-blankareas.md
```diff
@@ -0,0 +1,117 @@
...
+- Select an area, press `Del` key to remove it
+- Drag the start/end of an already existing area and move it along the line to reduce/enlarge it
+- Drag the zone entirely and move it along the line
+
+the map tool defined zone would be set on `QgsTemplatedLineSymbolLayerBase` symbol layers as data defined properties. If no data defined property has been set, a window would ask user if he wants to generate an auxiliary storage layer and which field he wants to use to join.
```
nyalldawson, line 75: Please use similar logic to the label tools to determine whether a field is available to store the values in -- there is nice logic there which handles expression based data defined values, eg `case when @map_scale > 20000 then "label_x_zoomed_out" else "label_x_zoomed_in" end` and extract that actual target field from that.

qep-345-line-symbollayer-blankareas.md
troopa81: No there is nothing missing. I just wanted to bring some context to explain that user can already use annotations to place marker wherever he wants. So this QEP is only about blank areas, nothing related to annotation.

qep-345-line-symbollayer-blankareas.md
troopa81: I'm still hesitating here between
- string comma separated
- add a new QgsProperty DataTypeNumericList in this [enum](https://github.com/qgis/QGIS/blob/a8f44fa00e7daa3c8c73d098d602c16ff76051ba/src/core/qgsproperty.h#L82)

With a preference for the second one. But I wanted to wait implementation to see what would be best. Do you have any opinion on this ?

qep-345-line-symbollayer-blankareas.md
troopa81: > This mockup is rather misleading in my opinion -- the list of regions will only be visible and editable IF it's a static list of regions which apply to ALL features. This is a very limited use case!

I considered this and was on the edge of removing the table widget completely. I kept it because:
- to keep consistent with all other data defined property where you can (always ?) give a static value
- I imagine user could define kind of meter defined dash pattern common to all lines. I'm really not sure this is really needed :man_shrugging:  


> in which case the regions WON'T be visible or editable in this list.

Like other data defined property, user could still use the widget to edit the default value, when no data have been defined. But they wouldn't be editable with the map tool which is quite disturbing I admit.


Maybe it's best to completely get rid of the widget and if user wants a static value, he can still use a static expression.

qep-345-line-symbollayer-blankareas.md
troopa81: I fail to find track of what you're mentioning. Where does that take place exactly?

qep-345-line-symbollayer-blankareas.md
nyalldawson: Can you drop this bit? I think it's misleading to include if you're not actually going to implement any changes there. 

(FWIW, I don't think annotations are a good solution for manual hash/marker placement -- there'd be no way to link the symbology of the original line to the custom annotation marker/hashes, so any need to change colors/line thickness/marker symbol would be a horrible painful experience requiring changes to many individual annotation objects as well as the original layer :scream: )

qep-345-line-symbollayer-blankareas.md
nyalldawson: > Shall we really ? Isn't Qt already avoiding drawing anything if you draw outside the device (image, widget) ? Do we have something to gain to check beforehand ?

I'm not sure if Qt optimises and skips work for raster paint devices, but at least for svg/pdf devices it definitely includes objects which are outside of the device area.

qep-345-line-symbollayer-blankareas.md
nyalldawson: I think list is theoretically better... but then we'd have patchy format support for it (ie not shapefiles). So unless there's fallback logic to use strings if the layer doesn't support lists, then I'd just play it safe and use strings alone.

qep-345-line-symbollayer-blankareas.md
nyalldawson: Yeah -- for development simplicity I would lean toward just skipping the widget. I don't think it'll end up being used much/at all.

qep-345-line-symbollayer-blankareas.md
troopa81: > (FWIW, I don't think annotations are a good solution for manual hash/marker placement -- there'd be no way to link the symbology of the original line to the custom annotation marker/hashes, so any need to change colors/line thickness/marker symbol would be a horrible painful experience requiring changes to many individual annotation objects as well as the original layer 😱 )

True. 

I replaced this bit with 

> User could then use another point layer with marker symbol to place and rotate other marker where he wants to. 

I don't want to add too much specific to marker line symbol layer (like adding, moving) and would like to rely on what already existing.

qep-345-line-symbollayer-blankareas.md
troopa81: OK, let's stick for string for now and I'll see if I list can be done (and if it's a good idea) during implementation

qep-345-line-symbollayer-blankareas.md
troopa81: Done 5e6a265

qep-345-line-symbollayer-blankareas.md
troopa81: Done aa15c1d

troopa81:
Thank you for review @nyalldawson , It's very much appreciated

qep-345-line-symbollayer-blankareas.md
signedav: Since this QEP is about blank areas only, this is out of scope, but still out of curiosity: What difference does it make when it's a point layer? It has neither a link to the original geometry, right?

benoitdm-oslandia:
+1

nyalldawson:
Declaring as passed with approval from @signedav , @benoitdm-oslandia , @nyalldawson

PR #347, opened 2025-08-26
Added to README.md a section "Navigating QEPs"

velle:
I was in doubt myself, which QEPs were in effect. 

E.g. some QEPs were listed in the README.md, while others were not, so I wondered if only the ones listed were in effect. 

So I added to README.md, a section on how "Navigating QEPs".

Helps reader understand:
- what is approved (and what is not)
- what is withdrawn (nothing yet)
- how to research for discussions on QEP (including finding draft QEPs that were rejected)

But please make sure that what I have written is correct before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,34 @@ When the discussion and voting period ends, one of the following actions are tak
   outcome of the proposal.
 - **If the proposal was rejected**: the Pull Request should be closed
 
-## List of approved QEPs
+## Navigating QEPs
+
+### Effective QEPs
+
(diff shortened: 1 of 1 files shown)


troopa81 (commented):
@velle Thanks for your contribution!

I'm concerned that what you are currently describing is just the natural process of GitHub repository and pull requests, and so, doesn't need to be describe again.

Again, not sure it needs to be described.

As for withdraw QEPs, I imagine that we would just remove the file. It would still be in git history if needed but would not bother people reading the repository anymore

PR #356, opened 2025-12-01
QEP 406: Markers/Hashes line symbol layer extra items

troopa81:
This is a follow up of [QEP #345](./qep-345-line-symbollayer-blankareas.md) which provides a tool to define per feature blank segments on marker/hashed line symbol layer where no markers/hashes are displayed (see [here](https://github.com/qgis/QGIS/pull/64047#issue-3659188974) for demo).

This QEP aims at providing tools to create, move, rotate and delete per feature new markers/hashes for marker/hashed line symbol layer. Those new markers will share the same style as the original markers/hashes, the same selected or visibly state. 

You can read it [here](https://github.com/qgis/QGIS-Enhancement-Proposals/blob/b18c8ff97239461d448abc74b66f861ba7c546bd/qep-406-markers-hashes-lines-extra-items.md)

**Funded by Stadt Frankfurt am Main and Oslandia**

PR #359, opened 2026-01-08
QEP 407: Introduce a data viewer for the model designer

ValentinBuira:
This QEP aims to bring a new integrated data viewer directly inside the model designer. This would allow to inspect and debug the state of your model as you build them.

Please see: `qep-407-dataviewer-modeldesigner.md`


nyalldawson:
Sounds exciting!

What about the performance issues with the attribute table? Showing an attribute table could potentially block the main thread for a loooooong time. We'd either need to explicitly limit the table to eg the first 200 rows, or make it hidden by default.

nyalldawson:
Follow up question - how will ownership of layers be handled?

troopa81:
Nice proposal.

Be careful that there is already a QEP [#406](https://github.com/qgis/QGIS-Enhancement-Proposals/pull/356)

Can you increment your proposal to 407 ?

ValentinBuira:
> If pagination requires the rewrite of the attribute page, I can totally understand that it is not feasible currently and this shouldn't block your proposal.

Yes it's outside the scope of this proposal, but make this QEP works with the current limitation of the attribute table was an import part of the proposal

cbertelli:
I think that one theme is the number of lines. I understand pagination is a problem, but having 50 lines as an example could be sufficient, finding out that the list stops at 200 could be difficult to undestand/accept.
This could be enhanced by a diagnostic tab, finding out duplicates, missing data and other frequent quirks of geographic data.

3nids:
this QEP has been up for voting since February 23rd, with a reminder on April 7th. 

I would have preferred avoid voting on this QEP personally but due to the lack of feedback, I am voting +1 now.

nyalldawson:
+1

nyalldawson:
Declaring as passed with votes from @3nids, @nyalldawson

Issue #350, opened 2025-09-29
Clarify how QEP numbers are to be assigned


strk:
The README.md file in the root of this repository documents the process of proposing a new QEP like this:

> make a copy of the QEP-0-Template.md file, populate the template and rename as required

But how to find out what kind of rename is required ? In other words, what number should be used for the new QEP file ? (and also which filename template, given most files have a lowercase "qep" prefix while the template and QEP4 is uppercase ?

PR #360, opened 2026-01-14
QEP 413: port TopoViewer to Browser

strk:
This would be one more step toward demoting of DBManager ( https://lists.osgeo.org/pipermail/qgis-developer/2022-June/064850.html ).

The reason why this QEP does't yet have a number at this stage is reported in #350


strk:
I've assigned this QEP a number and included some more implementation details, do you think more is needed ?

PR #363, opened 2026-01-31
Add QEP 408: AI tool use policy

rouault:
# Summary

LLM/AI are nowadays a reality, that whether we like it or not, introduces new
challenges in our collaborative environment that need to be addressed.

## Proposed Solution

The QEP 408 .md is added as a policy document in this repository

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ When the discussion and voting period ends, one of the following actions are tak
 - [#320 Backport policies](https://github.com/qgis/QGIS-Enhancement-Proposals/blob/master/qep-320-backport-policy.md). This
   QEP documents the standards and requirements for backporting changes from the ``master`` branch to Long Term Release (LTR)
   and current stable release branches in QGIS. Developers must follow these guidelines when submitting backports.
+- [#408 AI tool use policy](https://github.com/qgis/QGIS-Enhancement-Proposals/blob/master/qep-408-ai-tool-policy.md). This QEP documents the QGIS project policy regarding the use of AI/LLM tools by contributors.
 
 ## FAQ
 
--- a/qep-408-ai-tool-policy.md
+++ b/qep-408-ai-tool-policy.md
@@ -4,7 +4,7 @@
 
 **Author** Even Rouault (@rouault)
 
-**Contact** even at rouault at spatialys dot com
+**Contact** even dot rouault at spatialys dot com
 
 **Version** N/A
 
--- a/qep-408-ai-tool-policy.md
+++ b/qep-408-ai-tool-policy.md
@@ -1,0 +1,51 @@
+# QGIS Enhancement: AI tool use policy
+
+**Date** 2026/01/31
+
+**Author** Even Rouault (@rouault)
+
+**Contact** even at rouault at spatialys dot com
+
+**Version** N/A
+
+# Summary
+
+LLM/AI are nowadays a reality, that whether we like it or not, introduces new
+challenges in our collaborative environment that need to be addressed.
+
+## Proposed Solution
+
+A new page is added in the QGIS developer documentation. Its proposed content
+is at https://github.com/qgis/QGIS-Documentation/pull/10714
+
+## Deliverables
+
+New page in the QGIS developer documentation
+
+### Example(s)
+
+N/A
+
+### Affected files
+
+N/A
+
+## Risks
+
+None identified
+
+## Performance Implications
+
+N/A
+
+## Further Considerations/Improvements
+
+None
+
+## Backwards Compatibility
+
+N/A
+
+## Issue Tracking ID(s)
+
+N/A
--- a/qep-408-ai-tool-policy.md
+++ b/qep-408-ai-tool-policy.md
@@ -14,6 +14,13 @@ LLM/AI are nowadays a reality, that whether we like it or not, introduces new
 challenges in our collaborative environment that need to be addressed. This document
 describes the policy of the QGIS project in that matter.
 
+# Warning
+
+It is not currently a settled question within the QGIS community whether code
+created by LLMs is acceptable in the QGIS codebase at all. This QEP is addressed
+at the interaction aspects of LLM content, and should not be viewed as the project
+taking a position on whether or not LLM contributions are acceptable.
+
 # Policy
 
 QGIS's policy is that contributors can use whatever tools they would like to
--- a/qep-408-ai-tool-policy.md
+++ b/qep-408-ai-tool-policy.md
@@ -121,6 +121,8 @@ to help other reviewers prioritize their review time.
 
 If a contributor fails to make their change meaningfully less extractive,
 maintainers may lock the conversation and/or close the pull request/issue/QEP.
+In case of repeated violations of our policy, the QGIS project reserves itself
+the right to ban temporarily or definitely the infringing person.
 
 # Copyright
 
--- a/qep-408-ai-tool-policy.md
+++ b/qep-408-ai-tool-policy.md
@@ -11,41 +11,133 @@
 # Summary
 
 LLM/AI are nowadays a reality, that whether we like it or not, introduces new
-challenges in our collaborative environment that need to be addressed.
+challenges in our collaborative environment that need to be addressed. This document
+describes the policy of the QGIS project in that matter.
 
-## Proposed Solution
+# Policy
 
-A new page is added in the QGIS developer documentation. Its proposed content
-is at https://github.com/qgis/QGIS-Documentation/pull/10714
+QGIS's policy is that contributors can use whatever tools they would like to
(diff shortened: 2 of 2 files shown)


saberraz:
How about Plugins? I think there should be a more restrictive rules around plugins.

timlinux:
Well that is a really balanced and well thought out approach to LLM contributions, I would be happy to see that adopted as standard practice. Thank you for taking the initiative.

saberraz:
> > How about Plugins? I think there should be a more restrictive rules around plugins.
> 
> why more restrictive ? and what degree of control do we have in practice on plugins given they only get basic scrutiny?

Because it is usually less rigorous reviewed as the main QGIS code and can cause security issues if distributed through the QGIS plugin repo.

haubourg:
We can object that the maintenance burden is on the plugin's author's shoulders. If he decides to rely on LLM for this, this is his responsibility. 
For core it is a community matter

rouault:
I've moved the content of the QGIS documentation pull request back to the .md of this pull request, and integrated a remark about  possible ban of infringing contributors in case of repeated violations

nyalldawson:
@rouault
> that might be a bit extreme? 

That's why I'd put the qualifier "and are deceptive about". I.e. it's not that they didn't initially add the label, but rather if they've LIED when asked directly whether they've tested work or it was written by ai, then the one-strike rule kicks in. (Eg the stories told by https://github.com/qgis/QGIS/pull/64713#issuecomment-3805959273 / https://github.com/qgis/QGIS/pull/64722#issuecomment-3808770413 / https://github.com/qgis/QGIS/pull/64722#issuecomment-3808817136  or https://github.com/qgis/QGIS/pull/64701#issuecomment-3801601882 / https://github.com/qgis/QGIS/pull/64701#issuecomment-3810384631)

timlinux:
> As there are different approaches to that topic, I've created a poll at https://docs.google.com/forms/d/e/1FAIpQLSdnVWoD5DrwCbNXqPqHsLw2jfbLkPMKBkvfyQfTQOPZkj_EaQ/viewform . Please fill in

@rouault  Could we put your poll into the QGIS feed to get broader feedback?

rouault:
> Could we put your poll into the QGIS feed to get broader feedback?

up to you. The the target audience I had in mind was more QGIS devs & PSC as it is mostly a process/legal matter. But this indeed also affects users w.r.t bug reporting

elpaso:
> > > Could we put your poll into the QGIS feed to get broader feedback?
> > 
> > 
> > up to you. The the target audience I had in mind was more QGIS devs & PSC as it is mostly a process/legal matter. But this indeed also affects users w.r.t bug reporting
> 
> Yes I was more concerned that we ask these questions in an echo chamber which may not represent the broader sentiment of our users.
> 

I think that in this case we are more interested to the sentiment of our developers not to that of our users.

nyalldawson:
@rouault this has passed the 2 week discussion period now -- did you want to move it to voting in its current form?

gdt:
(I didn't answer the poll because google forms, but I think the policy is fine *if* qgis.org accepts LLM output at all.)

The CRS report is interesting.   The first part is focused on copyright registration, which I believe is required before enforcement.  The second part basically says that it is really unclear where the fair use boundary is, while being pretty clear by implication that absent fair use LLMs are simply infringing.   What it doesn't talk about at all is the question of whether the LLM model itself, that is trained on scraped intputs, is infringing.  It seems to only talk about the LLM outputs.

As for liability for infringing coyrights (held by those whose data was stolen for training), I don't see any basis to assign a particular entity as responsible and exclude others.   Any entity that does something reserved to the copyright holder could have some trouble.   So that could be the submitter, the merge approver, github (for now, until the non-free forge problem is fixed :-), qgis.org, and any downstream redistributors and even those who reproduce qgis internally without "distributing".   Probably qgis.org signed an agreement with github to indemify them against any claims "related to or arising from" or some such very broad language, and indemnification language doesn't necessarily require showing wrongdoing.   But it's very unclear as there are a lot of novel and unsettled questions; my view is that basic fairness dictates that  liability should mostly rest with LLM companies and secondarily with those submitting LLM output to projects.  Besides actually being liable, there's the difficulty of being wrongly hassled.   Plus there's the ethical issue; accepting LLM output is a decision that the process of training models on content scraped without permission, is morally acceptable.

As for LLVM, there's a corporate boundary where the LLVM foundation is a 501(c)3 (charitable non-profit under US tax code).   I have not generally had the impression that donors to a charity have any liability exposure (absent some conspiracy to set up a fake charity to do something intentionally and clearly unlawful).   So their risk is really "might be some judgement that llvm is infringing" with a tiny chance of "can't use llvm any more".    I wouldn't conclude from seeing sponsors that they have thought *at all* about risks to open source license integrity from LLM outputs.

To me the really big question is:  Are we going to
  - use a DCO, or some other clear mechanism like an inbound=outbound statement, so that contributors assert that they hold copyright or have the right to license the contribution under the project's ilcense?
  - drop any notion of contributors saying *anything* about whether they have the right right to license a contribution?
  - explicitly say that qgis.org has decided that LLM output does not infringe copyright so that's an exception to a DCO-like process

As far as I can tell, the qgis contribution/DCO situation is very hazy, at least from https://docs.qgis.org/3.40/en/docs/developers_guide/index.html

vpicavet:
Thanks @gdt I agree with you on all points.

Some additional views : 
- I don't think the survey respondant have a clear view of IP problems
- the report indeed tends to indicate that all LLM output could be public domain. I think that this would prevent us from applying a GPL licence on it  ( IANAL though ). They also say that this may depend on cases
- The report gives some interesting views, but it is not a legal document with legal precedence value. The main conclusion of the report is that law should be adapted to handle these cases
- the report concerns US law. Since the Berne convention, copyright laws are supposed to be coherent and compatible, but on this specific topic, it may vary a lot. Also, this area changes every single week with new legal cases opened
- How could QGIS.org cannot "decide" that LLM output does not infringe copyright ? This is not something we can just decide
- as for all legal stuff, when lawyers are asked to play in a grey zone, their answer is usually : do not get involved, period. Risk is something lawyers do not like at all.

And again, even with a DCO putting a clear mechanism of inbound=outbound responsibility, I find it extremely dangerous, and even unethical : 
- the risk of being sued (QGIS.org, mergers, distributors…) and having to deal with legal problems stays the same, even if the final culprit is the contributor, the mess to deal with the legal aspect of things is still there
- I am OK to push responsibility to contributors IF and only IF they are given the tools and competences to assess and prove this responsibility. Here, we force responsibility, but there is no way for a contributor to assess IP compliance for LLM-generated code. This would be, coming from an opensource community, very harsh as a position

Most probably, we would need legal professional to help QGIS.org here.

rouault:
> Did you want to move this to voting stage now?

yes, all, please cast your votes.

Starting with my +1

nyalldawson:
@wonder-sk @haubourg @timlinux @PeterPetrik @alexbruy @troopa81 @elpaso @nirvn @gacarrillor Heads up that you'll need to re-cast your previous +1 votes after the policy wording was changed.

uclaros:
+1

gacarrillor:
+1, thanks again for this!

3nids:
+1

elpaso:
+1

lindangulopez:
Thanks @nyalldawson and everyone involved!

I look forward to seeing how this evolves alongside legal and technical discussions around AI in open source projects.

I’m reaching out to get your input on RSE community needs for AI/LLM metadata and to ensure that proposed fields **baseModel, agentRole, humanOversight, aiAssistance, and reviewedByHuman** are considered for inclusion in **CodeMeta v4**: see discussion [here](https://github.com/codemeta/codemeta/discussions/460).

What do you think about integrating these minimal AI metadata fields? Adding **[CodeMeta-JLD](https://github.com/codemeta/codemeta/blob/master/codemeta.json)** and a QGIS [CodeMeta-crosswalk](https://github.com/codemeta/codemeta/tree/master/crosswalks) could be bonus.

Kind regards,
Linda

haubourg:
> I’m reaching out to get your input on RSE community needs for AI/LLM metadata and to ensure that proposed fields **baseModel, agentRole, humanOversight, aiAssistance, and reviewedByHuman** are considered for inclusion in **CodeMeta v4**: see discussion [here](https://github.com/codemeta/codemeta/discussions/460).
> 
> What do you think about integrating these minimal AI metadata fields? Adding **[CodeMeta-JLD](https://github.com/codemeta/codemeta/blob/master/codemeta.json)** and a QGIS [CodeMeta-crosswalk](https://github.com/codemeta/codemeta/tree/master/crosswalks) could be bonus.

This would deserve a dedicated discussion, I'm not totally aware of the benefits, and we might need to learn about all this.

PR #364, opened 2026-02-01
QEP 409: Attribute Form QML Widget Editing Capabilities

nirvn:
_This QEP proposes to add editing capabilities for attributes form QML editor widgets._

Since 2018, users have gained the ability to add QML widgets to enhance their attribute forms. These widgets have so far been read-only and used to display feature attributes through QML scenes utilizing QML items such as graphs.

While this is a great addition when using feature forms to read features, the QML widgets could be even more useful if they would allow for feature attribute editing. 

Proposed solution and benefits in attached QEP.

### DIFF
--- a/qep-409-attribute-form-qml-widget-editing.md
+++ b/qep-409-attribute-form-qml-widget-editing.md
@@ -1,0 +1,85 @@
+# QGIS Enhancement: Attribute Form QML Widget Editing Capabilities
+
+**Date** 2026/02/01
+
+**Author** Mathieu Pellerin (@nirvn)
+
+**Contact** mathieu@opengis.ch
+
+**Version** QGIS 4.2
+
+# Summary
+
+Since 2018, users have gained the ability to add QML widgets to enhance their
+attribute forms. These widgets have so far been read-only and used to display
+feature attributes through QML scenes utilizing QML items such as graphs.
+
+While this is a great addition when using feature forms to read features, the
+QML widgets could be even more useful if they would allow for feature
+attribute editing. 
+
+## Proposed Solution and Benefits
+
+This QEP proposes the introduction of a new object accessible within the QML
+widget’s scene with invokable functions to act as a bridge between the scene
+within QML widget and its attribute form. The newly introduced object – injected
+into the scene’s root context - will allow for accessing the attribute form
+context and enable attribute editing of the attribute form’s current feature.
+
+This additional ability for QML widgets can unlock a world of new possibilities
+by allowing for QML components to drive the creation of new user interface to
+manipulate feature attributes.
+
+### API Considerations
+
+For QML widgets to be able to edit the attribute form’s current feature, the
+`QgsWidgetWrapper` class will need to be tweaked to add a signal to handle
+attribute value changes. 
+
+As for the QML widget itself, a new `QgsAttributeFormQmlWidgetBridge` object
+will be injected into the root context of the QML scene that will offer a few
+invokable functions and a context property to drive attribute editing. 
+
+```
+class QgsAttributeFormQmlWidgetBridge : public QObject
+{
+    Q_OBJECT
+
+    Q_PROPERTY( QgsAttributeFormContext context READ context NOTIFY contextChanged )
+
+  public:
+    Form() = default;
+    ~Form() = default;
+
+    QgsAttributeFormContext context() const;
+    void setContext( const QgsAttributeFormContext &context );
+
+    Q_INVOKABLE setAttribute( const QString &name, const QVariant &value );
+}
+```
+
+The object itself will be able to be expanded in the future to add more
+functionalities if need be.
+
+Note that the `QgsAttributeFormContext` class is already a `Q_GADGET`, which
+means the only work required on that front will be to expose some of its
+properties as `Q_PROPERTY` for QML widgets to be able to receive contextual
+awareness of its current mode, the current feature, the parent feature, etc.
+
+## Deliverables
+
+A new  `QgsAttributeFormQmlWidgetBridge` class which will act as the bridge
+between the scene within the QML widget and its attribute form.
+
+### Affected Files
+
+- qgsattributeform.cpp
+- qgsattributeformcontext.h
+- qgswidgetwrapper.cpp
+- qgswidgetwrapper.h
+- qgsqmlwidgetwrapper.cpp
+- qgsqmlwidgetwrapper.h
+
+## Risks
+
+None
--- a/qep-409-attribute-form-qml-widget-editing.md
+++ b/qep-409-attribute-form-qml-widget-editing.md
@@ -71,6 +71,56 @@ means the only work required on that front will be to expose some of its
 properties as `Q_PROPERTY` for QML widgets to be able to receive contextual
 awareness of its current mode, the current feature, the parent feature, etc.
 
+### Sample QML editor widget code
+
+This is a sample QML editor widget code that provides an example of how the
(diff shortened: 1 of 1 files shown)


qep-409-attribute-form-qml-widget-editing.md
```diff
@@ -0,0 +1,85 @@
...
+As for the QML widget itself, a new `QgsAttributeFormQmlWidgetBridge` object
+will be injected into the root context of the QML scene that will offer a few
+invokable functions and a context property to drive attribute editing. 
+
+```
+class QgsAttributeFormQmlWidgetBridge : public QObject
+{
+    Q_OBJECT
+
+    Q_PROPERTY( QgsAttributeFormContext context READ context NOTIFY contextChanged )
```
signedav:
```suggestion
    Q_PROPERTY( QgsAttributeEditorContext context READ context NOTIFY contextChanged )
```
It's `QgsAttributeEditorContext` or am I mixing something up?

qep-409-attribute-form-qml-widget-editing.md
```diff
@@ -0,0 +1,85 @@
...
+    void setContext( const QgsAttributeFormContext &context );
+
+    Q_INVOKABLE setAttribute( const QString &name, const QVariant &value );
+}
+```
+
+The object itself will be able to be expanded in the future to add more
+functionalities if need be.
+
+Note that the `QgsAttributeFormContext` class is already a `Q_GADGET`, which
```
signedav: Will it be able to provide default value and constraint information?

nirvn: @signedav , it'd be possible yes. Not part of the initial implementation scope though.

qep-409-attribute-form-qml-widget-editing.md
nirvn: Correct. My word dyslexia hitting again :)

qep-409-attribute-form-qml-widget-editing.md
```diff
@@ -0,0 +1,85 @@
...
+```
+class QgsAttributeFormQmlWidgetBridge : public QObject
+{
+    Q_OBJECT
+
+    Q_PROPERTY( QgsAttributeEditorContext context READ context NOTIFY contextChanged )
+
+  public:
+    Form() = default;
+    ~Form() = default;
```
wonder-sk: nitpick: wrong constructor/destructor names

qep-409-attribute-form-qml-widget-editing.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+The object itself will be able to be expanded in the future to add more
+functionalities if need be.
+
+Note that the `QgsAttributeEditorContext` class is already a `Q_GADGET`, which
+means the only work required on that front will be to expose some of its
+properties as `Q_PROPERTY` for QML widgets to be able to receive contextual
+awareness of its current mode, the current feature, the parent feature, etc.
+
```
wonder-sk, line 73: I am trying to wrap my head around how the QgsAttributeFormQmlWidgetBridge usage would look in QML - a tiny sample QML widget with editing would be great 🙂

qep-409-attribute-form-qml-widget-editing.md
```diff
@@ -0,0 +1,85 @@
...
+manipulate feature attributes.
+
+### API Considerations
+
+For QML widgets to be able to edit the attribute form’s current feature, the
+`QgsWidgetWrapper` class will need to be tweaked to add a signal to handle
+attribute value changes. 
+
+As for the QML widget itself, a new `QgsAttributeFormQmlWidgetBridge` object
+will be injected into the root context of the QML scene that will offer a few
```
wonder-sk, line 40: Isn't injection to the root context of QML considered a legacy practice in Qt6, due to trouble with tooling etc? Wouldn't it be better that the QML widgets would always include a _required_ property "context", which would be provided when creating the widget?

nirvn: @wonder-sk , I don't think it's anywhere close to being deprecated (not like function parameter injection). The documentation however does note it should be avoided, TIL.

I'd be -0 on going for a property context. We can easily replace that with a singleton (i.e. qmlRegisterSingletonType). It'll get us to the same end. Would you be satisfied with that?

qep-409-attribute-form-qml-widget-editing.md
nirvn: @wonder-sk , here's a tiny sample showing how the bridge would be used to set an attribute.

```
import QtQuick
import QtQuick.Controls
import QtQuick.Layouts

/**
 * Note: QgsAttributeFormQmlWidgetBridge exposed as a singleton named attributeFormBridge
 */ 

Item {
  width:400
  height: childrenRect.height
  
  ColumnLayout {
    width: parent.width
    
    Label {
      id: label
      // Example of dynamic binding using the formFeature property of the context
      text: "Current 'field_a' value" + attributeFormBridge.context.formFeature.attribute("field_a")
    }
    
    Button {
      id: button1
      text: "Set field attribute value to 'one'"
      
      onClicked: {
        // Using the bridge to set a field/attribute value on the current feature
        attributeFormBridge.setAttribute("field_a", "one");
      }
    }
    Button {
      id: button2
      text: "Set field attribute value to 'two'"
      
      onClicked: {
        // Using the bridge to set a field/attribute value on the current feature
        attributeFormBridge.setAttribute("field_a", "two");
      }
    }
  }
}
```

Does that help?

wonder-sk: That helps, thanks! 👍  (it may be useful to have the code sample in the QEP content too)

Side note: how about adding `attributeFormBridge.attribute()` getter method as a shortcut, for consistency with setAttribute()?

qep-409-attribute-form-qml-widget-editing.md
wonder-sk: I'm easy here - I'm no expert on QML, so I am happy with whatever you are happy 😄  (I just read about these "newer" code patterns recently, so I was curious whether they make sense here)

qep-409-attribute-form-qml-widget-editing.md
nirvn: > That helps, thanks! 👍 (it may be useful to have the code sample in the QEP content too)

OK, I'll add.

> Side note: how about adding attributeFormBridge.attribute() getter method as a shortcut, for consistency with setAttribute()?

@wonder-sk , yep, I added this and well, forgot to push the change. QEP updated now.

nirvn: @wonder-sk , sample code added into the QEP. Thanks your engagement and your suggestions :pray:

qep-409-attribute-form-qml-widget-editing.md
```diff
@@ -0,0 +1,140 @@
...
+  width:400
+  height: childrenRect.height
+  
+  ColumnLayout {
+    width: parent.width
+    
+    Label {
+      id: label
+      // Example of dynamic binding using the formFeature property of the context
+      text: "Current 'field_a' value" + attributeFormBridge.context.formFeature.attribute("field_a")
```
wonder-sk, line 99:
```suggestion
      text: "Current 'field_a' value" + attributeFormBridge.attribute("field_a")
```

Thanks for the updates!

nirvn: @wonder-sk , actually there I think it's more useful to demonstrate the bindable context property, don't you think?

PR #367, opened 2026-03-10
QEP 412: 3d editing support

benoitdm-oslandia:
This QEP proposes introducing vector layer editing directly within the 3D view by adding dedicated 3D user interface tools. 

Here are the user stories for each feature described in the QEP:

* Active Layer Selection from the 3D View
  As a user working in the full-screen 3D view, I want to be able to switch the active layer using a selector integrated into the 3D view, so I don't have to switch to the main 2D window.
* Creating 3D Primitives
  As a GIS user, I want to be able to create 3D shapes (cube, sphere, torus, etc.) directly in the 3D view by entering their parameters with the mouse or keyboard, in order to quickly add geometric entities to my PolyhedralSurface vector layer.
* 3D Boolean Operations (Intersection, Difference, Union)
  As a GIS user, I want to be able to apply Boolean operations (intersection, difference, union) between several selected 3D entities, in order to generate new complex geometries from existing objects.
* Move / Rotate / Scale 3D Entities
  As a GIS user, I want to be able to move, rotate, or resize vector entities directly in the 3D view using the mouse or keyboard, with axis constraints, to precisely adjust their position and shape.
* 3D Mirroring
  As a GIS user, I want to be able to mirror a 3D entity along a standard plane (XY, XZ, YZ) or an arbitrary plane defined by three points, to quickly create symmetrical geometries.
* Duplicate 3D Entities
  As a GIS user, I want to be able to duplicate one or more 3D entities by defining a direction, distance, and number of copies, to efficiently reproduce objects in the 3D scene.
* 3D Splitting
  As a GIS user, I want to be able to split a PolyhedralSurface entity along a plane (standard or arbitrary) to obtain two new, distinct geometric entities.
* 3D Extrusion
  As a GIS user, I want to be able to extrude faces, polygons, or linestrings from a PolyhedralSurface entity by defining a direction and a height to model 3D volumes from existing geometries.

cc @ptitjano, @lbartoletti, @signedav


dvdkon:
Nitpick: You probably want ["editing"](https://en.wiktionary.org/wiki/editing), not ["edition"](https://en.wiktionary.org/wiki/edition).

haubourg:
Hi, we discussed QEP process in last PSC and agreed that for QEP that imply end user changes would deserve a two stage QEP: 
- user stories in the main description, so that stakeholders, power users, and overloaded PSC benevolent can discuss the approach 
- the QEP files for implementations details, powerfull for change tracking and advanced technical discussions

As this QEP probably introduce something massively new for users, could you elaborate the user stories, ergonomical choices in the QEP description ?

qep-412-3d-primitive-edition.md
```diff
@@ -0,0 +1,238 @@
...
+
+* does the toolbar should be activated for the layer?  `bool accept( QgsMapLayer *layer )`
+* activate the toolbar `void activate( QgsMapLayer *layer )`
+* deactivate the toolbar `void deactivate()`
+
+### Active layer selection
+
+Currently, to select the active layer, a user must open the QGIS 2D main window and choose a layer in the layer browser panel. However, when working in a 3D view, the user may maximise it for easier navigation. To change the active layer in this situation, the user has to lower the 3D view, find the QGIS 2D main window, select a layer, and then return to the 3D view. This process is very inefficient.
+
+This could be solved by adding a drop-down selector with only the editable and VectorLayer layers. This drop-down selector will be synchronized with the one on the 2D view. When a layer is selected in this drop-down selector, its editing mode is enabled.
```
signedav: Maybe it is out of scope here, but would it make sense to make the 3D view dockable in future and use it in the same way as the canvas? Or are there logical restrictions I do not see?

qep-412-3d-primitive-edition.md
```diff
@@ -0,0 +1,238 @@
...
+
+### 3D arrangement operations
+
+This proposal plans to add the following arrangement operations to the toolbar:
+
+* move
+* rotate
+* scale
+* mirror
+* copy   =============================== array
```
signedav: ?

qep-412-3d-primitive-edition.md
```diff
@@ -0,0 +1,238 @@
...
+* first, the user have to select 1 entity
+  * entity can be selected using the 2D main window by any available method
+  * entity can also be selected in the 3D view by picking them on screen, using Shift/Ctrl modifiers to add or remove entities from the selection
+* If an arbitrary plane is chosen, it must be defined by selecting three points in the 3D view
+* the user validates the operation using the validate button in the dialog or by pressing the ENTER key
+* since the map tool is still active, the user can apply the same operation again
+
+#### Copy
+
+This operation creates one or multiple features, here is its user workflow:
```
signedav, line 150: Would a simple copy+paste+move be possible as well?
If no, I think I would preferably name this here duplicate instead of copy. But just an opinion...

qep-412-3d-primitive-edition.md
benoitdm-oslandia: the 3D view is already dockable but the use case here is when the 3D view is full screen. But it can be removed.

qep-412-3d-primitive-edition.md
benoitdm-oslandia: missed that one

qep-412-3d-primitive-edition.md
benoitdm-oslandia: ok

qep-412-3d-primitive-edition.md
uclaros: I suppose what David means is to make it dockable over the main 2d map canvas and have some tab selector to switch between 2d map canvas and 3d map canvas, similar to how arcgis works.
This would eliminate the need for changing the active map layer by means other than the layers panel.
It would also allow for working on the 3d map canvas with ctrl+tab and ctrl+shift+tab, which will be super nice!

haubourg:
> Hi, we discussed QEP process in last PSC and agreed that for QEP that imply end user changes would deserve a two stage QEP:
> 
>     * user stories in the main description, so that stakeholders, power users, and overloaded PSC benevolent can discuss the approach
> 
>     * the QEP files for implementations details, powerfull for change tracking and advanced technical discussions
> 
> 
> As this QEP probably introduce something massively new for users, could you elaborate the user stories, ergonomical choices in the QEP description ?

Thanks a lot Benoit

qep-412-3d-primitive-edition.md
benoitdm-oslandia: Yes I should be nice but it is not in the scope of this QEP :) !

qep-412-3d-primitive-edition.md
```diff
@@ -0,0 +1,238 @@
...
+
+The new map tools will be grouped in 4 new toolbars to increase the modularity the `Qgs3DMapCanvasWidget` class. These toolbars will keep their map tools outside the `Qgs3DMapCanvasWidget` class which in return will handle a bunch of editing toolbars. To do so a new abstract class `Qgs3DEditingToolBar` will be added to help in the creation of these editing toolbars.
+
+Currently, the point cloud editing code is directly included in `Qgs3DMapCanvasWidget`. Using the same approach for the four new classes would make the code difficult to maintain. Therefore, the following changes are proposed:
+
+* Move the existing point cloud editing code into its own file.
+* Introduce an abstract class `Qgs3DEditingToolBar`, which all editing toolbars will inherit from, and move the point cloud editing code into this new API.
+* Introduce four new editing toolbars, each inheriting from `Qgs3DEditingToolBar`.
+
+`Qgs3DEditingToolBar` should have this API:
```
uclaros, line 42: How do you plan the four toolbars to be activated and interact with each other? `Qgs3DMapCanvasWidget` is not a `QMainWindow` and the toolbars do not behave like proper QToolbars. They are practically just container widgets spanning the whole dialog's width and they cannot be moved/reordered/stacked like normal toolbars. Activating 4 toolbars at once will cover a big part of the canvas.

qep-412-3d-primitive-edition.md
```diff
@@ -0,0 +1,238 @@
...
+
+* does the toolbar should be activated for the layer?  `bool accept( QgsMapLayer *layer )`
+* activate the toolbar `void activate( QgsMapLayer *layer )`
+* deactivate the toolbar `void deactivate()`
+
+### Active layer selection
+
+Currently, to select the active layer, a user must open the QGIS 2D main window and choose a layer in the layer browser panel. However, when working in a 3D view, the user may maximise it for easier navigation. To change the active layer in this situation, the user has to lower the 3D view, find the QGIS 2D main window, select a layer, and then return to the 3D view. This process is very inefficient.
+
+This could be solved by adding a drop-down selector with only the editable and VectorLayer layers. This drop-down selector will be synchronized with the one on the 2D view. When a layer is selected in this drop-down selector, its editing mode is enabled.
```
uclaros: When implementing point cloud editing, we decided not to add a layer selector in the 3d view. The layers panel is the central place that handles the current layer. Users know how to identify their layers in the layers panel and it is proven than when the layers are presented in another form (eg combobox in processing tools) users start having difficulties finding the desired layer.
If it feels like the layers panel is too far from the 3d view (like the maximized 3d view example) then the right way to go is to try to bring the 3d view closer to the main window, like @signedav suggested and not the other way round. This is the paradigm that's used in arcgis pro too.

Automatically turning editing mode on is something that does not fit well with how QGIS works so far. I believe that explicitly toggling editing through a pencil icon should be a required step before modifying data.

qep-412-3d-primitive-edition.md
```diff
@@ -0,0 +1,238 @@
...
+
+### 3D primitive creation
+
+These new map tools will need the SFCGAL library to be efficient (at least v2.3.0).
```
uclaros, line 50: Do you plan to have 2 implementations, one efficient using SFCGAL and one inefficient without SFCGAL?

qep-412-3d-primitive-edition.md
```diff
@@ -0,0 +1,238 @@
...
+#### Move, rotate, scale
+
+These operations modify the existing features and share the same user workflow:
+
+* select the active layer to edit (any QgsVectorLayer). The new features will be saved in this layer
+* click on the edit button of the 3D view, the 3D arrangement operation toolbar is enabled
+* click a button to select the operation map tool, the map tool user input dialog is displayed
+* first, the user have to select at least 1 entity
+  * entities can be selected using the 2D main window by any available method
+  * entities can also be selected in the 3D view by picking them on screen, using Shift/Ctrl modifiers to add or remove entities from the selection
```
uclaros: How will entities get picked on 3d view? There is no raycasting support for point and line layers.

qep-412-3d-primitive-edition.md
```diff
@@ -0,0 +1,238 @@
...
+
+#### Move, rotate, scale
+
+These operations modify the existing features and share the same user workflow:
+
+* select the active layer to edit (any QgsVectorLayer). The new features will be saved in this layer
+* click on the edit button of the 3D view, the 3D arrangement operation toolbar is enabled
+* click a button to select the operation map tool, the map tool user input dialog is displayed
+* first, the user have to select at least 1 entity
+  * entities can be selected using the 2D main window by any available method
```
uclaros: Selected features may not be 3d or may not have a 3d renderer. How will those be treated?

images/qep412/create_cube_p3.png
uclaros, line 1: What does the rubber band represent? Shouldn't the mouse follow the geometry borders?

uclaros (commented):
It seems that this proposal is mostly about editing polyhedral surface layers in 3d. Arguably, most gis data consist of points/lines/polygons and most QGIS users will expect editing in 3d views to support digitizing and basic manipulation of such layers and not creating a torus in full 3d.

Maybe I'm old and narrow minded, but I consider editing of polyhedral surfaces to be the domain of 3d modeling software and not gis. 3D modeling software are beasts of their own, their developers have spent countless hours dealing with details and making them usable. However much effort we put in replicating 3d modeling flows in QGIS, we will never reach the usability of dedicated 3d modeling tools. Hence my suggestion is to limit the scope of 3d editing to basic digitizing, move, scale and rotate tools.

The rest can be implemented in a plugin until the usability is proven and the UX is settled. We can focus in exposing to python any parts of the API that would be needed to achieve that.

One alternative that I'd also be happy with is what Gemini suggests that works in Arcgis pro (I haven't seen it in action but the description makes sense): A user can select a polyhedral surface to be edited in external editor. Object is exported to .glb file and blender or other software is started, allowing editing of the exported file. When the file is saved, the geometry is automatically updated in gis 3dview and data provider.

qep-412-3d-primitive-edition.md
benoitdm-oslandia: Toolbars can also be presented as drop down menu like image. But we can also think to rework this window to be a real QWindow and add proper QToolbar support (but not on this QEP).

qep-412-3d-primitive-edition.md
benoitdm-oslandia: Ok for the layer panel: we will not add a drop-down selector. 

But about the pencil icon, I think the "bubble" icon (see below) can be replaced by the pencil to display the proper editing toolbar (3D primitive, point cloud, etc.) and to toggle the layer edit mode as soon the toolbar is shown. Also the greyed pencil icon on the image below is redundant with the fact the toolbar is displayed and can be removed

image

qep-412-3d-primitive-edition.md
benoitdm-oslandia: Only one implementation will be provided but alternative could be implemented in future QEPs.

qep-412-3d-primitive-edition.md
benoitdm-oslandia: a POC of raytracing support has been made which work with model point and could be extended to billboard. Line has not been investigated but could work the same way as point.

images/qep412/create_cube_p3.png
benoitdm-oslandia: The rubber does not represent something useful at this stage, without constraint it is more a way to remember the number of parameter currently set: 3 segments <==> 2 parameters set and working on the 3rd. If the axis constraint is applied, it makes more senses as the rubberband points are aligned to the primitive.

qep-412-3d-primitive-edition.md
benoitdm-oslandia: good point! As it may create confusing workflow in the 2D view, we will only select 3D objects from the 3D view.

benoitdm-oslandia:
> Maybe I'm old and narrow minded, but I consider editing of polyhedral surfaces to be the domain of 3d modeling software and not gis. 3D modeling software are beasts of their own, their developers have spent countless hours dealing with details and making them usable. However much effort we put in replicating 3d modeling flows in QGIS, we will never reach the usability of dedicated 3d modeling tools. Hence my suggestion is to limit the scope of 3d editing to basic digitizing, move, scale and rotate tools.

I agree we are not creating a CAD software but only proposing some basic tools. And these tools can include primitive creation and editing as they can help in improving scene content (like solar panel implementation or street equipment for examples). Also like the current point cloud editing toolbar, the user may need to quickly edit a Polyhedral surface which was badly digitalized.

benoitdm-oslandia:
Hi all!  Can you, please cast your votes?

qep-412-3d-primitive-edition.md
uclaros: Then what does `need the SFCGAL library to be efficient` mean?

qep-412-3d-primitive-edition.md
uclaros: > Ok for the layer panel: we will not add a drop-down selector.

It is still in the proposal though.

qep-412-3d-primitive-edition.md
uclaros: It is not clear to me what you will eventually do about this.

uclaros (commented):
I still consider 3d boolean operations and arbitrary primitive creation _not_ being basic tools for a gis. Basic tools would be point, line, polygon creation, vertex and feature moving, rotation. I think having the former without the later is like putting the cart before the horse.

rouault:
+1

kikislater:
+1

qep-412-3d-primitive-edition.md
benoitdm-oslandia: > Toolbars can also be presented as drop down menu like image.

So the space will be limited to 4 drop down menus.

qep-412-3d-primitive-edition.md
benoitdm-oslandia: indeed ... fixing

qep-412-3d-primitive-edition.md
benoitdm-oslandia: Sorry but I do not understand what you did not understand :grimacing: : there is currently (except with SFCGAL) no efficient way to do the tasks implied, so we will need SFCGAL or any new/future implementation that does the same job. And as I said above, we will only implement SFCGAL bindings to do the jobs.

wonder-sk:
-1 in the current state

(I apologize for the late feedback, it was on my to-do list to review but couldn’t get time)

# The problem

This is my central assumption, around which the argumentation follows:
**For consistent user experience, 3D tools should work the same (or at least similar) to 2D tools.**

We should replicate tool behaviors that users are already familiar with. I would like to argue that this QEP ignores much of that. I do not like the idea of doing the same kind of operation with different behavior just because we are in 3D view and not in 2D view. I hope we agree on that 🙂 

Here’s a comparison of the tools proposed in the QEP and their 2D counterparts:

- create primitive (cube/sphere)
    - 2D tools: “Rectangle from 3 points” (and friends), “Circle by a center point and another point” (and friends, including ellipse). User clicks few points and confirms with right click.
    - proposed 3D tools: a dialog pops up(!), then user picks a point (is it a centroid?), then there’s setup of the primitive size involving Tab/Shift+Tab and clicking or numerically entering data, followed by confirmation by clicking a button or pressing Enter.
    - my commentary: this is COMPLETELY different workflow, even though I do not see a fundamental reason for it. I understand there’s the idea of allowing numerical input of parameters, but that should be handled by a 3D equivalent of “advanced digitizing panel”, rather than replicating some of its functionality across each tool

- 3D boolean operators
    - 2D tools: “Merge selected features” for union, no equivalent for intersection/difference. User first picks features with a selection tool, then click the button
    - proposed 3D tools: user clicks features with Shift/Ctrl modifiers, then confirms with Enter to create new feature
    - my commentary:
        - for union and intersection, one could simply use vector layer’s selection rather than having to pick features in the tool. The difference operation is special as it depends on the order, so there’s no way around it.
        - I do not get why Shift/Ctrl modifiers are needed though, wouldn’t just clicking features be enough (clicking an already picked feature would deselect it).
        - I would expect that the original features would get replaced by the new feature - otherwise this kind of construction of geometry with boolean ops will require removal of temporary geometries all the time.

- move / rotate / scae tools
    - 2D tools: “Move Feature”, “Rotate Feature”, “Scale Feature”. User clicks a feature, moves/rotates/scales it and clicks again to confirm. For rotation/scale, QGIS shows user input widget in top-right corner for numerical input if needed.
    - proposed 3D tools: dialog pops up(!), then user clicks one feature (or more, using shfit/ctrl), moves mouse (or edits some value(?) in the dialog), confirms with Enter
    - my commentary:
        - it is unclear what the input dialog for the tools would contain. If it is rotation angle / scale factor, that makes sense, if there’s also e.g. translation distance when moving, that’s again territory of the “advanced digitizing panel”
        - I don’t understand why user should be picking multiple features with shift/ctrl when we could simply use vector layer’s selection, like the 2D tools already use.
        - Also unclear why the operation should be confirmed with Enter or clicking a button, when it is natural to just click at the new location

- duplicate tool
    - 2D tools: “Copy and Move Feature(s)”. User clicks a feature, and with every next click it adds a duplicate at a new position. Right click terminates the tool. “Copy Features in an Array” is another one, more advanced, to create a predefined number of features.
    - proposed 3D tools: dialog pops up(!), then user clicks one feature (or more, using shift/ctrl), then sets number of copies, confirms with button/Enter
    - my commentary:
        - it looks like the idea is to implement something like the advanced “copy features in an array” 2D tool (which I didn’t know about until today). Unless strictly needed by the client, I would actually recommend to skip this tool altogether and do the duplicate functionality by letting user do copy+paste feature and then use the move tool.

- split tool
    - 2D tool: “Split Features”. User clicks two or more points to create a linestring, confirms with right click, a feature is split.
    - proposed 3D tool: dialog pops up, then user clicks one feature (or more using shift/ctrl), then user somehow picks three points in case of arbitrary plane (initiated from the dialog?), then the plane may be moved (with mouse or editing plane parameters)
    - my commentary:
        - we can’t have the simplicity of 2D tool as we’re working with full plane and not just some “split polygon” so we need to pick features instead of using all that would be intersected. But we should use the current selection instead of custom picking of features at the tool start.
        - having three points as a way to define a plane makes sense, but that is going to be very difficult for users! Other 3D software like Blender allow you to draw a line in 3D view and it will be extended to produce cutting plane (see e.g. https://www.youtube.com/watch?v=ZYYkdNhfMhw).
        - the QEP split will create one new feature (and modify the existing one?). From my understanding, multiple new features may be produced (e.g. when cutting U-shaped geometry), because a polyhedral surface geometry must not have multiple parts.
        - how will the split work - does it keep the polyhedral surface hollow inside, or will it try to create a watertight object by adding a new face to the cut geometries?

- mirror tool
    - 2D tools: none (available in 3rd party QGIS plugins)
    - proposed 3D tool: similar to split tool
    - my commentary:
        - split tool was meant to allow editing the selected plane after picking three points, but mirror tool does not seem to allow that? that is not consistent
        - I would suggest to skip mirror tool from the initial implementation - it looks like very advanced tool (given that there’s no such 2D tool)

- extrude tool
    - 2D tools: no equivalent
    - proposed 3D tool: user picks one or more faces (again with shift/ctrl modifiers), then moves the mouse to set extrusion height, clicks button/Enter to confirm
    - my commentary:
        - face selection: like for earlier tools, not sure why shift/ctrl modifiers are needed to select/deselect faces?
        - how would it work with multiple selected faces - each face would be extruded perpendicularly to it, or the tool would pick a single extrusion direction for all?

# Random other notes

My understanding is that conceptually, the modelling approach outlined by this QEP is to start from some basic shapes, and then apply boolean operations as the main way to build more complex geometries (CSG style). While that is a legitimate approach and used widely in CAD (e.g. for mechanical parts), I wonder why this was chosen compared to the approach centered around face/edge/vertex level editing (which GIS users are more comfortable with). The “vertex tool” for 3D editing will be missing, as well as creation of simple point/line/polygon in 3D view. Simple things like slightly moving an existing wall require significant effort without a “vertex tool” - essentially to create a new cube, rotate, scale and position it properly, then subtract it (or union).

I feel strongly about the fact that the 3D editing tools should use the existing concept of selected features - there’s no reason not to do that - it would simplify the tools and make them consistent with 2D tools. There should be one 3D map tool that would do the selection, operating consistently with how the default 2D map tool for selection works.

There is no tool to remove objects - are users expected to remove them in 2D canvas? Especially if boolean operations will be creating new features, there will be a strong need to delete intermediate features.

Primitive creation: users can create a torus (which does not even have a 2D tool), but the more basic primitives are missing:
- rectangle, polygon - those are basic valid polyhedral surfaces useful for further modelling
- prism or pyramid - for roofs

Usage of dialogs for editing tools: 2D editing tools extensively use the QgsUserInputWidget, which neatly sticks to the top-right corner of canvas. Typically it does not need any buttons to confim an action, with some exceptions like simplify tool. I would strongly suggest to have something similar for 3D views, rather than normal dialogs that may pop up and cover the canvas.

Regarding the 3D equivalent of “advanced digitizing panel” - my suggestion would be to initially skip the numerical input of parameters for the 3D editing tools, and focus on the usability of the operations. The only bits that are required is the ability to constrain a change (e.g. move/rotate) to a single axis or a plane. Other bits like specifying absolute or relative coordinates, that is something that can be added afterwards (like it happened with 2D map tools).

How about cut/copy/paste support? This has not been mentioned.

How about snapping? It is not mentioned anywhere whether it is considered and how would it work.

Primitive creation - looking at screenshots for cube/torus: I do not understand why “Scale” transformation is needed when user has explicit edit widgets to set the primitive’s dimensions.

Globe: are you going to support 3D editing in globe mode? Currently the vector layer support for globe is poor, but that should be improved soon (it is within our last year’s crowdfunding’s scope).

3D boolean operations: if I understand correctly, these operations work with triangular meshes. Does that mean that a polyhedral surface (like a cube) will end up being a triangular mesh after a boolean operation, or is there some magic that will “fix” the geometry back to polyhedral surface where possible?

Have you considered how editing tools would interact with camera control? For example, measure tool or most of the point cloud editing tools allow you to move/rotate the scene even while doing input for the tool. Would this be supported in 3D vector editing tools too, or will there be some conflicts why this would not be possible?

I assume the 3D editing tools, in order to be practical, will need to have things like “zoom to feature” - is something like that planned?

Do we really need four new editing toolbars? If I count correctly, it should be 11 new buttons for the various tools, something that a single toolbar could handle easily. Also, I am not sure if there’s any need for Qgs3DEditingToolBar abstraction (but that’s too low level for this discussion).

# Sample editing cases

The proposal does not state the expected use cases, and how the added tools satisfy them. Use cases help to imagine the whole workflow, and identify where the user may struggle.

Let me try to add one - I want to create a model of a house, with a gable roof, something like that:

image

I assume these will be two features: one for the house, one for the roof, so they have different colors.

For the house itself, I guess I would build a cube and scale it. Then, I guess I would need to do either two splits to cut away the top part where the roof will be placed, or create two rotated cubes and use difference tool to remove unwanted material (the first approach is probably simpler).

Now for the roof - I think I would need one cube, scale and rotate it to fit one side of the roof, then duplicate it and either rotate it or mirror it, and finally merged them with the union operator. I guess placing the roof “cube” correctly may be a bit of a challenge, and require moving it from multiple side views to place correctly.

(Feel free to add more representative use cases!)

haubourg:
I'm late to the party too. 
From my old experiments when just discussing about the "clic and drag" to "clic and clic" UX changes, I tested a lot of CAD and 3D tools to have a better view of existing 3D editing tooling. I would have liked this proposal to compare the reference ergonomic approaches ( Sketchup vs AutoCad3D vs Fusion ) and explain why choosing one ergonomy instead of another one. And open the path to the possibility to change the desired way to handle 3D. 
We will not reinvent all this in QGIS and I think the best approach is to mimic existing ways. 

Back in those days, we concluded QGIS should never become a 3D editing tool. What has changed ?

nyalldawson:
> Back in those days, we concluded QGIS should never become a 3D editing tool. What has changed ?

Maybe because we now have a great inbuilt 3d data viewer? (Have you ever tried to MAKE 3d data in qgis? It's horrible to do, if not impossible. Last time I tried I had to resort to hand writing WKT strings for the data :sob: )

haubourg:
> Maybe because we now have a great inbuilt 3d data viewer? (Have you ever tried to MAKE 3d data in qgis? It's horrible to do, if not impossible. Last time I tried I had to resort to hand writing WKT strings for the data 😭 )

I don't doubt we miss the 3D edition feature. I just realize we have to discuss the UX choices at start at length, because we might end up in some sort of mixed ergonomy. But we can also state we open this as an experimental field, and accept massive changes from one release to another. I would be ok with this approach

signedav:
> I don't doubt we miss the 3D edition feature. I just realize we have to discuss the UX choices at start at length, because we might end up in some sort of mixed ergonomy. But we can also state we open this as an experimental field, and accept massive changes from one release to another. I would be ok with this approach

Yes. Since it's likely that the true quality of the UX will be visible once non-developers start working with it anyway, I would not see it as critical to introduce map tools that ain't intended to last forever. But I'm not very familiar with the 3D viewer UX topic to be able judge.

nyalldawson:
Given @wonder-sk 's -1, this proposal is blocked in its current form. I've downgraded it back to "in discussion" for further refinement.

wonder-sk:
Finally I'm back from the summer break (🏖️), sorry for the longer period of silence...

Let me try to address the latest proposal from Benoit:

> UI approach: User input widgets will be displayed in the 3D view's lateral panel rather than modal dialogs.

Input widgets in a lateral panel are certainly better than pop-up dialogs, but I still don't see a reason why the overall concept of creation of features is different from 2D canvas: I still think the creation should be primarily mouse-driven (like in 2D canvas) with a generic approach to enter exact values (like with advanced digitizing panel in 2D canvas - of course it does not need to be as complicated!). And there should not to be a special "Create" button to confirm - it should be driven by the mouse (just like in 2D canvas).

It would be really good to have the same kind of QgsUserInputWidget experience like in 2D canvas. Generally, the idea has been to get rid of the "navigation" panel for good - any debugging bits are gone now, and it just waits for a brave soul to move its functionality as an optional overlay directly into the 3D canvas...

> Object selection: Users select objects before activating a map tool (consistent with 2D map tools). Selection can happen in the 2D main window (standard tools) or the 3D view (ray tracing-based 3D picking). Boolean difference is a special case since it is not commutative: it may require a dedicated workflow

👍 

 > Deletion, copy, and paste tools will be added.

Great!

> Toolbar organization: The four toolbars could be consolidated, but some must operate on different layer types and appear contextually. For example, move/rotate/scale applies to both Polyhedral and Polygon layers, while primitive creation is exclusive to Polyhedral layers.

That's fine - some tools can simply stay disabled when their geometry type does not match - just like it is the case with 2D editing tools.

> Operation result handling: A checkbox will let users choose whether to keep the operation result for reuse elsewhere or replace the original object.

I don't have a strong opinion, so this could work, but maybe it would be neater to just always replace the original object (as that's likely the wanted scenario most of the time) and when user needs to reuse the object, they would copy it beforehand? (it could save us an extra checkbox)

> Primitives: No primitives are missing from the set.

A simple rectangular plane is missing in my opinion (because it is possible to combine them and then extrude at once for more complex building shapes) - but I can live without it 🙂 

---

I realize there are some unanswered questions from Benoit's earlier response - shall I try to answer to all of it? I fear we'll get into too many parallel discussion threads at once. Also generally, it is easier for me to give feedback on any proposed design ideas, rather than providing alternative designs (which is hard and takes a lot of time). So I wonder how to best move the discussion forward - would a call be useful? Or even better, if you're heading to the QGIS contributor meeting, that will be a great opportunity to discuss things in a lot of detail...

signedav:
> And there should not to be a special "Create" button to confirm 

I like that. Exactly what we see in Benoit's video on 1:00 m but instead of "create" one can press righ mouse key, right?
And would this be possible @benoitdm-oslandia ?

PR #372, opened 2026-04-02
QEP 415 Refactor the evaluation of processing model with a dependency graph

ValentinBuira:
This QEP highlights some limitations with the internal data structure representation of the model in QGIS. And proposes a series of refactor based on results from graph theory and compiler theory, to _ensure the continuous improvement of the model designer_, and _bring some performance gains on completion of this QEP_.

With at the backbone of this proposal, the introduction of a dependency graph. 

See file : `qep-415-dependancy-graph-processing-model.md`

This QEP will be submit as part of the [QGIS.org grant program](https://blog.qgis.org/2026/03/16/qgis-grants-11-call-for-grant-proposals-2026/)


qep-415-dependancy-graph-processing-model.md
```diff
@@ -0,0 +1,192 @@
...
++    QMap< QString, QList< QString > > mChildAlgorithmGraph;                                                                                       
++      
+```
+
+
+This introduction of the DAG will trickle down (in a good way) to some methods related to dependencies:
+
+| Method  | Expected changes |
+| ------------- | ------------- |
+| `dependentChildAlgorithmsRecursive`  | All descendants of algorithms using standard BFS. instead of current implementation Mark parameter  QString &conditionalBranch as unused   |
```
troopa81, line 86:
```suggestion
| `dependentChildAlgorithmsRecursive`  | All descendants of algorithms using standard BFS ( [Breadth-first_search](https://en.wikipedia.org/wiki/Breadth-first_search) ) . instead of current implementation Mark parameter  QString &conditionalBranch as unused   |
```

> Mark parameter QString &conditionalBranch as unused

Can you elaborate ? Are talking about the `branch` argument ? Why don't we need it anymore ?

qep-415-dependancy-graph-processing-model.md
```diff
@@ -0,0 +1,192 @@
...
+
+```
+While there is child algorithms not executed 
+     Loop for each child algorithm not yet executed
+            If all dependencies of the child algorithm are executed:
+                    Execute algorithm and mark it as executed
+            // if it’s a conditional branch prune child algorithm not yet executed when branch is falsy
+```
+
+In current implementation, every child algorithm is checked at each iteration, and we hope for the best that they are somehow sorted in the right order. In worst case scenario this as an algorithm with polynomial complexity where n is the number of child algorithm
```
troopa81, line 122: Wouldn't that be negligible compare to the execution of the task themselves ? Mainly because QGIS model doesn't contain a huge amount of tasks. Even with 40 tasks, iterating over the task list every time a task is finished is certainly not ideal, but does that have a substancial impact on the overall model running time ?

Did you manage to measure that impact in some way?

qep-415-dependancy-graph-processing-model.md
ValentinBuira: Yes, so if we look at the usage of this argument, it's only used in the pruning branch logic in the execution of the model ( in processAlgorithm at https://github.com/qgis/QGIS/blob/fc7ae590e963cdc2314c5f35544419afafd9e6c8/src/core/processing/models/qgsprocessingmodelalgorithm.cpp#L626) (to note it pass through `dependentChildAlgorithms` first but this call `dependentChildAlgorithmsRecursive`)

As with the Khan's algorithm we won't need the pruning during the execution because with the graph traversal these branch just would not being visited.

Making this argument effectively obsolete, but just keep it around as unused as it's exposed through the python API

qep-415-dependancy-graph-processing-model.md
ValentinBuira: > Wouldn't that be negligible compare to the execution of the task themselves ? 

Your intuition might be correct that in terms of performance it would be marginal in the execution of the model. But the execution of the model is not the only place where it would beneficial to have a graph traversal for exemple _dependentChildAlgorithmsRecursive_ that we talked above also made needless loop at each recursion steps 

Moreover the bulk of performance improvements will need the introduction of the dependency graph at some point. for example the execution of the child algorithm is still sequential today and could be parallelized, and more optimiziation are possible with graph, quoted from the QEP:  


>   *  Performance, execution speed
> 
> kahn’s algorithm offers already better performance than the current implementation, but the kahn’s algorithm can also be adapted to be parallelized to compute independent branches concurrently.
> 
> Schutte, J. (2013). Implementing Parallel Topological Sort in a Java Graph Library [Master's thesis, University of Twente].
> 
> Moreover, once we have a DAG data structure, a plethora of algorithms and optimizations are available to us. For example: transitive graph reductions, Sub tasking, disjoint path for concurrent execution
> 
> Slim Ben-Amor, Liliana Cucu-Grosjean. Graph reductions and partitioning heuristics for multicore DAG scheduling. Journal of Systems Architecture, 2022, 124, pp.102359. 10.1016/j.sysarc.2021.102359 . hal-03538053


**To me the introduction of a dependency graph is the *sine qua none condition* to have all of these**

m-kuhn:
I've seen (large) models eat up all memory in real world projects, having a proper architecture to deal with that is much needed.

qep-415-dependancy-graph-processing-model.md
```diff
@@ -0,0 +1,214 @@
...
+        remove edge e from the graph
+        if m has no other incoming edges then
+             insert m into ChildAlgSet
+    ++ flush temporary results from n
+
+Now in the context of the model designer for debugging purposes it can be useful to keep temporary output, but we could imagine a toggle, to favor debug or performance.  
+
+* Fix export to python 
+
+Export a model to a python implementation is borken as soon as there is a branch in the model (see ticket https://github.com/qgis/QGIS/issues/49053). Because the export to python flatten all algorithms and not necessarily in the right order
```
troopa81:
```suggestion
Export a model to a python implementation is broken as soon as there is a branch in the model (see ticket https://github.com/qgis/QGIS/issues/49053). Because the export to python flatten all algorithms and not necessarily in the right order
```

qep-415-dependancy-graph-processing-model.md
troopa81: > for example the execution of the child algorithm is still sequential today and could be parallelized, 

That could be done in current state, running all tasks eligible to be executed in parallel while all are executed.

But I agree that it helps reducing other issues and would improve, even if it is only slightly, performance.

qep-415-dependancy-graph-processing-model.md
troopa81: Thanks for the explanation, maybe it's worth completing the QEP with for later reviewer.

troopa81 (commented):
What if, someday, we want to add a `for loop` element to iterate over a task ? Would it keep being an acyclic graph? Would it be easy to deal with?

ValentinBuira:
> What if, someday, we want to add a `for loop` element to iterate over a task ? Would it keep being an acyclic graph? Would it be easy to deal with?

This is hardly a simple answer, because there is a lot of "it depends" on how we implement it, but I'm glad you asked, and generally speaking I think so and I had this use case in mind.

In an initial first draft of this proposal I had a second part, using [visitor pattern](https://craftinginterpreters.com/representing-code.html#the-visitor-pattern) to evaluate the model. 

The visitor pattern would make easier to change how the model is traversed, and make it easy to add new type of node/statement 

with I believe the following additional key benefits:

* The visitor pattern would pave the way for more than simple model execution traversal but also real control flow such as iterator, if, else.
* Untied the execution from the graph traversal, with more flexibility to modify each one independently
 
## Comparison with interpreted language

Interpreted programming language is also an area where you take a serie of linear statement(text line) and introduce control flow when you execute the program. 

### Traditional pipeline 
Usually the pipeline  in an interpreted language look like this ( this is simplified for the example)

`Text` -> `Parse into concrete syntax tree` -> `construct Abstract Syntax Tree` -> `Evaluate AST with visitor pattern`

Example in python with the [AST module](https://docs.python.org/3/library/ast.html) where you can reimplement `ast.NodeVisitor` with a def `visit_For` when traversing a for loop. Where you visit the body of the loop,  the condition, handle variable scope etc... 

### Hypothetical pipeline in QGIS

In the context of QGIS the model is already a structured so we don't need to parse a text file 
`Model` -> `Construct AST / or Directed acyclic graph` -> `Evaluate AST/Dag with visitor pattern`

*The directed acyclic graph acts would act as the AST.* 


So in rough pseudo code for visitor pattern in QGIS we would  have 

```
class CORE_EXPORT NodeVisitor:
{
 public:
   virtual ~NodeVisitor() = default;
   virtual void visitChildAlgorithm( const QgsProcessingModelChildAlgorithm *modelAlgorithm ) = 0;
   virtual void visitConditionalBranch( const QgsProcessingModelChildAlgorithm *modelAlgorithm ) = 0;
   virtual void visitLoop( const QgsProcessingModelChildAlgorithm *modelAlgorithm ) = 0; // Here we can add a new loop without disrupte the rest of the tree traversal
   // Add more as needed
}
```

And the model itself would override each type of node we have
```
class CORE_EXPORT QgsProcessingModelAlgorithm : public QgsProcessingAlgorithm, public NodeVisitor
{
```

<br>
So to answer the question : 

> Would it keep being an acyclic graph? 

It would still be a directed acyclic graph but the traversal would be different

But I did not add in in the final proposal, because I wanted to keep things simple. And I am still not 100% sure if a DAG can be easily substituted with an  AST.

to note:  it's not uncommon for compiler and interpreter to have multiple intermediate representation (AST, control flow graph, etc...)  and back and forth between them 


*Just to make it clear nothing from this message is binding in the proposal*

anitagraser:

nyalldawson:
@ValentinBuira can you update the proposal text to reflect the discussion outcomes before we start lodging votes on this?

ValentinBuira:
> @ValentinBuira can you update the proposal text to reflect the discussion outcomes before we start lodging votes on this?

I updated the proposal. The proposal didn't changes much, I just added some clarification on the raised points and I added a paragraph in the risk section on the development process that was already detailed in the google forms. 

It is now in a state ready for voting

PR #373, opened 2026-04-02
QEP 416: Expression filter on vector layers

signedav:
This QEP proposes a new filter option on vector layers similar to the query builder, accessible via *Layer > Filter*.

This enables us to filter all types of vector layers according to QGIS expressions and use elements such as variables, custom functions and more.

See details in the description [here](https://github.com/signedav/QGIS-Enhancement-Proposals/blob/expression-filter/qep-416-expression-filter.md).

- [ ] Formulate solution for project scope


signedav:
Can we already vote on that? There was no feedback here, but since I discussed the implementation beforehand with @nyalldawson and @m-kuhn, I think the proposed solution is solid.

nyalldawson:
@signedav can you give me a day or so to read over? I haven't had a chance yet

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+
+**Author** Dave Signer (@signedav)
+
+**Contact** david at opengis dot ch
+
+**Version** QGIS 4.2
+
+# Summary
+
+An expression filter similar to the query builder, accessible via *Layer > Filter*, where the entire layer can be filtered using expressions. You can choose between one of the filters. This enables us to filter all types of vector layers according to QGIS expressions and use elements such as variables, custom functions and more.
```
elpaso, line 13: do you mean QGIS expressions?
Also, it isn't clear to me what do you mean with "You can choose between one of the filters." ?

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+  
+  [...]
+
+  request.combineFilterExpression(source.subsetExpression(), true);
+
+  [...]
+
+```
+
+`combineFilterExpression` merges the expression with the existing filter epression using `AND`. The `combineFilterExpression` should be extended by a parameter `bool outer = false` which we set to `true` to place the `subsetExpression` before the `filterExpression` (becoming the outer filter). This is due to short-circuit evaluation (returning directly when the first expression is `false`) and the assumption that the `subsetExpression` is generally less complex than the `filterExpression`. Of course, this is not always the case. Perhaps a cost-calculator would make sense, but this is not part of this enhancement.
```
elpaso, line 63: I think it's preferable to create a separate method, perhaps `prependFilterExpression()` ?

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+
+Also, it needs to set the filter type to `Expression` because the filter expression will not be evaluated on the client side otherwise.
+
+```c++
+  request.setFilterType( Qgis::FeatureRequestFilterType::Expression);
+``` 
+
+(`setFilterType` does not exist yet)
```
elpaso, line 71: Looking at the current implementation, the filter type is set implicitly by the constructor, and it's not supposed to be changed by client code (that's why there is no setter method for the filter type).

Have you considered an alternative approach where we keep the current distinction between the provider-level "subsetString" and the (temporary) vector layer-level filter expression and you transparently change the behavior of the provider-level subsetString to handle expressions instead? 

I find it less intrusive, the required changes should be minimal if the expression can be compiled, if not, you could transparently handle the filtering iterating the features with the expression at the provider level.
 
I believe that a check for the compatibility of the expression with a server-side compiled expression would be useful either ways to help the users to assess the performance issues they would face by using an expression that cannot be fully evaluated server-side.

elpaso (commented):
This is essentially a core implementation of https://plugins.qgis.org/plugins/qgis-subset-expression-plugin/ , right?

qep-416-expression-filter.md
signedav: Yes, QGIS expressions.

Between SQL and QGIS Expression filter. See the mockup:

image

qep-416-expression-filter.md
signedav: You mean to avoid having the `outer` parameter? Would be fine for me as well.

qep-416-expression-filter.md
signedav: > Looking at the current implementation, the filter type is set implicitly by the constructor, and it's not supposed to be changed by client code (that's why there is no setter method for the filter type).

Actually this is not needed. I've just seen that with `combineFilterExpression` does `setFilterExpression` and there it changes the `mFilter` already to `Qgis::FeatureRequestFilterType::Expression`. So no new setter needed.

> Have you considered an alternative approach where we keep the current distinction between the provider-level "subsetString" and the (temporary) vector layer-level filter expression and you transparently change the behavior of the provider-level subsetString to handle expressions instead?

So you mean that it is possible to enter QGIS expressions in the current Query Builder and they will be recognized? If so, I think that is fault-prone because users will then start mixing the two syntax. Or do you mean purely in the implementation and the user is still shown two modes?

> I believe that a check for the compatibility of the expression with a server-side compiled expression would be useful either ways to help the users to assess the performance issues they would face by using an expression that cannot be fully evaluated server-side.

Yes. What I suggested here https://github.com/qgis/QGIS-Enhancement-Proposals/pull/373/changes#diff-250c3b8149ea4fbb8c52746c084c6aa781bca8a2634bd1b154c9c473b9d19e8aR145 but it should not be a mandatory part of the first implementation.

signedav:
> This is essentially a core implementation of https://plugins.qgis.org/plugins/qgis-subset-expression-plugin/ , right?

Yes. Not a direct implementation of it but of its functionality. And I believe the initiating client is the same.

qep-416-expression-filter.md
elpaso: > > Have you considered an alternative approach where we keep the current distinction between the provider-level "subsetString" and the (temporary) vector layer-level filter expression and you transparently change the behavior of the provider-level subsetString to handle expressions instead?
> 
> So you mean that it is possible to enter QGIS expressions in the current Query Builder and they will be recognized? If so, I think that is fault-prone because users will then start mixing the two syntax. Or do you mean purely in the implementation and the user is still shown two modes?
> 

I think the dual-mode GUI is ok, what I think it's important is that once the filter is set, the client code and the API continue to work transparently without the need to know which type of filter is set.

Speaking of which, did you consider the possibility of setting both filters for the same source? (I mean combined with an AND). 

I don't know if that makes much sense but I can imagine that we could eventually desire a complex filter that on one end uses some specific server side functions AND on the other end some QGIS expressions that are not available server side.

nyalldawson: I personally don't see a compelling use case for both provider + qgis filters being active concurrently, and think it would add a lot of UI complexity vs a mutually exclusive approach.

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+
+When "WHERE clause" is chosen, you will see the "traditional" query builder content (Mockup):
+
+![querybuilder-gui](images/qep416/querybuilder-gui.png)
+
+When "QGIS Expression" is chosen, you will see the expression builder where you can create your expression (Mockup):
+
+![expression-gui](images/qep416/expression-gui.png)
+
+This would include at least project and global scopes.
```
nyalldawson: I'd only promise access to the global scope. While we could have additional scopes available via the feature request's expression context, we don't have any way to also make those scopes available elsewhere (eg to feature count, extent methods, etc). So we'd end up with an awkward situation where an expression filter may work or not depending on what method is called.

Also note that supporting project scope is non-trivial. We don't have access to the project from the provider (without the now banned QgsProject::instance() hack currently in use by some badly behaved providers). And there's also thread safety issues with accessing the project from background threads. (We do this a lot, it's bad, let's stop).

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+
+![querybuilder-gui](images/qep416/querybuilder-gui.png)
+
+When "QGIS Expression" is chosen, you will see the expression builder where you can create your expression (Mockup):
+
+![expression-gui](images/qep416/expression-gui.png)
+
+This would include at least project and global scopes. 
+
+We won't offer the combination of both. Even though it would be technically possible, it seems fairly fault-prone and hard to use. Using either one or the other seems right.
```
nyalldawson:
```suggestion
We won't offer the combination of both. Even though it would be technically possible, it seems fairly fault-prone and hard to use. Making the choice between provider side filter or QGIS expression filter mutually exclusive will lead to a more user-friendly experience. (If a user wants to combine a provider side filter with qgis expression filtering, they have other options like setting a rule-based renderer on the layer to apply a qgis expression filter at render time only.)
```

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Backend
+
+### Set the subset expression
+
+This subset expression will be set via the layer `QgsVectorLayer::setSubsetExpression()` and provider `QgsDataProvider::setSubsetExpression()`, eventually landing in the `QgsAbstractFeatureSource` as the member `mSubsetExpression`.
```
nyalldawson: This isn't clear. We need a way to reflect that subset expressions are mutually exclusive with subsetStrings. I'd lean toward deprecating the existing setSubsetString method and adding a new method which accepts a string filter and Qgis::ProviderFilterType enum instead.

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+
+```c++
+  request.setFilterType( Qgis::FeatureRequestFilterType::Expression);
+``` 
+
+(`setFilterType` does not exist yet)
+
+### Merging to the fids
+
+There is also a way to filter according to feature IDs: `QgsFeatureRequest( QgsFeatureId fid )` and `QgsFeatureRequest( const QgsFeatureIds &fids )`.
```
nyalldawson: Can you reword this to make it clear that you'll be adding this? Ie we need a way to upgrade an existing fid or fids type request to a filter expression type.

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+## Backend
+
+### Set the subset expression
+
+This subset expression will be set via the layer `QgsVectorLayer::setSubsetExpression()` and provider `QgsDataProvider::setSubsetExpression()`, eventually landing in the `QgsAbstractFeatureSource` as the member `mSubsetExpression`.
+
+It is stored as the custom property `QgsMapLayer::mCustomProperties['storedSubsetExpression']` (aligned with the `subsetString`).
+
+In the feature iterator, we then merge it into the feature request's filter expression so they can be compiled together into a server side filter (completely or partially) if possible. Otherwise, they are evaluated together on the client side.
+
```
nyalldawson, line 43: Can you add a note that support for this will need to be added provider-by-provider, so the existing provider capabilities metadata will be extended to add a flag to indicate if this support is available for a particular data provider?

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+  if ( request.filterType() == Qgis::FeatureRequestFilterType::Fid )
+  {
+    QString fidWhereClause = QgsPostgresUtils::whereClause( mRequest.filterFid(), mSource->mFields, mConn, mSource->mPrimaryKeyType, mSource->mPrimaryKeyAttrs, mSource->mShared );
+
+    whereClause = QgsPostgresUtils::andWhereClauses( whereClause, fidWhereClause );
+  }
+
+```
+
+So nothing has to be done here; it is just that it does not exclude an additional expression filter (the subset expression), which is done by setting the filter type as mentioned above.
```
nyalldawson: I disagree with this approach. I'd rather see a generic method added to QgsFeatureRequest which converts it from a Fid/FIDS type to a filter expression type, building the equivalent QGIS expression. Then we would make sure that providers which have support for qgis client side expression filters always can handle compiling those converted expressions.

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+  [...]
+```
+
+### When having no feature request filter yet
+
+On `Qgis::FeatureRequestFilterType::NoFilter`, the `subsetExpression` has to be added to the (empty) `filterExpression` and the filter type needs to be upgraded like above as well.
+
+### Additional places to consider
+
+#### Extend
```
nyalldawson:
```suggestion
#### Extent 
```

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+
+On `Qgis::FeatureRequestFilterType::NoFilter`, the `subsetExpression` has to be added to the (empty) `filterExpression` and the filter type needs to be upgraded like above as well.
+
+### Additional places to consider
+
+#### Extend 
+
+For example, via "Zoom to Layer(s)".
+
+This can be integrated once we are able to compile the subset expression.
```
nyalldawson: How?

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+### Affected Files
+
+- qgsvectorlayer.h / cpp to setSubsetExpression etc. and probably featureCount/extent
+- qgsdataprovider.h / cpp to setSubsetExpression etc. and probably featureCount/extent
+- qgsfeaturerequest.cpp to improve `combineFilterExpression` and add the subset expression to the `QgsAbstractFeatureSource`
+- Provider-specific feature iterators (like e.g. qgspostgresfeatureiterator.cpp)
+
+## Deliverables
+
+- Expression filter on all vector layers
```
nyalldawson: That'd imply adding support for this to ALL vector providers, which is a LOT of work. I'd explicitly list which providers you will do this for here.

qep-416-expression-filter.md
```diff
@@ -0,0 +1,181 @@
...
+
+Visual feedback would be very helpful in the GUI to know whether the expression can be *compiled* or not. But this is not part of this enhancement.
+
+### Affected Files
+
+- qgsvectorlayer.h / cpp to setSubsetExpression etc. and probably featureCount/extent
+- qgsdataprovider.h / cpp to setSubsetExpression etc. and probably featureCount/extent
+- qgsfeaturerequest.cpp to improve `combineFilterExpression` and add the subset expression to the `QgsAbstractFeatureSource`
+- Provider-specific feature iterators (like e.g. qgspostgresfeatureiterator.cpp)
+
```
nyalldawson, line 155: The provider compliance test suite should also be upgraded to perform tests on providers using qgis client side filters.

qep-416-expression-filter.md
signedav: > I think the dual-mode GUI is ok, what I think it's important is that once the filter is set, the client code and the API continue to work transparently without the need to know which type of filter is set.

Should `subsetString`, for example, return the query regardless of which filter is currently active? Perhaps we can ensure that without needing to store the expressions in the same variable. For instance, we could have new members like `subsetProviderString` and `subsetExpression`, where the relevant one is returned by `subsetString`. However, I have to think about that again.

> Speaking of which, did you consider the possibility of setting both filters for the same source? (I mean combined with an AND).

I tend to avoid that as well, since it is hard for a user to evaluate which filter is currently in use and why. I already wrote that here https://github.com/qgis/QGIS-Enhancement-Proposals/pull/373/changes#diff-250c3b8149ea4fbb8c52746c084c6aa781bca8a2634bd1b154c9c473b9d19e8aR31 and have now accepted Nyall's wording.

qep-416-expression-filter.md
signedav: Hm... I understand. However, the sponsor who came up with writing me this proposal original use case is that they can use project variables in the filter. And although a client requirement should not be directly related to the QEP, I would still need to think about a solution.

nyalldawson: Ok.. not saying it's impossible, but that's a good 2-4 days to get right on its own. 😱

qep-416-expression-filter.md
signedav: And store both into `subsetString`? Okay. I thought a little bit about whether we could keep the "inactive" filters in case the mode is changed, but it might be saver this way. This would also align with @elpaso 's  opinion, as far as I understand.

qep-416-expression-filter.md
signedav: Good to know. Thanks.

qep-416-expression-filter.md
signedav: What you suggest is primarily for uniformity and clearity, right? That would be okay with me.

qep-416-expression-filter.md
signedav: Okay, yes, maybe we should clarify this here. I would start with Postgres only and maybe OGR, but the reason I wrote "all vector layers" here is because the proposal is to have it for all of them one day. But  this may not be the right approach and I have to write that this can be achieved step-by-step...

qep-416-expression-filter.md
signedav: For e.g. Postgres in `computeExtent3D` by passing the compiled subset expression instead of the `filterWhereClause`. Or do you see a problem there? Otherwise, I will formulate it here in more detail.

PR #374, opened 2026-04-08
QEP 417: Replace SIP_FACTORY with std::unique_ptr

troopa81:
This QEP proposes to replace the use of SIP_FACTORY with std::unique_ptr returned type.

See details in the description [here](https://github.com/troopa81/QGIS-Enhancement-Proposals/blob/sip_factory/qep-417-remove-sip-factory.md).

### DIFF
--- a/qep-417-remove-sip-factory.md
+++ b/qep-417-remove-sip-factory.md
@@ -1,0 +1,71 @@
+# QEP 417: Replace SIP_FACTORY with std::unique_ptr
+
+**Date** 2026/04/08
+
+**Author** Julien Cabieces (@troopa81)
+
+**Contact** julien dot cabieces at oslandia dot com
+
+**Version** QGIS 4.2
+
+# Summary
+
+There are 2 types of methods returning an objet in QGIS API:
+- Method caller gets the ownership of the returned object (like all [clone()](https://github.com/qgis/QGIS/blob/d0e537f550fd433e67315ded6706f90cbeb1078d/src/core/geometry/qgscurve.h#L48) methods for instance). The caller is responsible to delete it, otherwise we would get a memory leak.
+- Method caller doesn't get the ownership of the returned object (like [QgsProject::mapLayer()](https://github.com/qgis/QGIS/blob/d0e537f550fd433e67315ded6706f90cbeb1078d/src/core/project/qgsproject.h#L1242) for instance). The callee object has still the ownership of the object (`QgsProject` owns `QgsMapLayer` in this example), and the caller **must not** delete it.
+
+This requires an explicit description so that Python, which manages memory differently than C++, knows precisely how to act in each scenario.
+
+QGIS API uses a special macro called SIP_FACTORY for methods which transfer the ownership of the returned object to the caller. This macro is translated as `/Factory/` SIP annotation which tells Python to delete the returned object when it goes out of scope.
+
+On the other hand, QGIS API sometime uses, rightfully, a `std::unique_ptr` as a return type instead of a raw pointer.
+
+Both `std::unique_ptr` and `SIP_FACTORY` describes the same paradigm, making the second redundant. Moreover, QGIS application could benefit from wider use of `std::unique_ptr` to improve memory management.
+
+
+## Proposed Solution
+
+This QEP proposes to: 
+- Modify `sipify.py` script to add the `/Factory/` for every methods whom return type is a `std::unique_ptr`
+- Use a `std::unique_ptr` as a return type for every methods which has the `SIP_FACTORY` and remove the use of `SIP_FACTORY` for these methods
+- When no methods are using the `SIP_FACTORY` anymore, remove completely the `SIP_FACTORY` definition
+
+As a consequence, all methods already returning a `std::unique_ptr` but missing `SIP_FACTORY` (like [QgsLineSymbol::createSymbol](https://github.com/qgis/QGIS/blob/master/src/core/symbology/qgslinesymbol.h#L36) for instance) would be automatically fixed. At the moment, there are more than 100 of these in QGIS API, **all leading to memory leak when called from a Python context**. 
+
+## How to port the code
+
+There are actually 1700 uses of `SIP_FACTORY`. I'm planning on using a Clang plugin to port the easy ones ([here](https://github.com/qgis/QGIS/blob/d0e537f550fd433e67315ded6706f90cbeb1078d/src/gui/symbology/qgssymbollayerwidget.h#L553) for instance). For the more complex ones, and probably for all those methods calls, rewrite would require to be done manually.
+
+## Deliverables
+
+Pull requests to replace raw pointer returned type with `std::unique_ptr`, and removing `SIP_FACTORY`.
+
+### Example(s)
+
+[Example on QgsProcessingModelAlgorithm](https://github.com/qgis/QGIS/commit/d0e537f550fd433e67315ded6706f90cbeb1078d)
+
+### Affected Files
+
+- All header files containing `SIP_FACTORY`, and the associated *.sip.in*
+- `sipify.py`
+
+## Risks
+
+Achieving such migration is complex and we could lack time to complete it. Worst case scenario would be that we fail to migrate all methods, but that would still be a better situation than what we have now.
+
+## Performance Implications
+
+None
+
+## Further Considerations/Improvements
+
+This would be a joined effort with the on going [Security project for QGIS](https://oslandia.com/en/security-project-for-qgis/) and the already proposed [pull requests](https://github.com/qgis/QGIS/pulls?q=is%3Apr+%22Use+unique_ptr+when+class%22+is%3Aclosed) aiming at improving memory safety.
+
+
+## Backwards Compatibility
+
+This is completely backwards compatible.
+
+## Issue Tracking ID(s)
+
+None
--- a/qep-417-remove-sip-factory.md
+++ b/qep-417-remove-sip-factory.md
@@ -25,12 +25,11 @@ Both `std::unique_ptr` and `SIP_FACTORY` describes the same paradigm, making the
 
 ## Proposed Solution
 
-This QEP proposes to: 
-- Modify `sipify.py` script to add the `/Factory/` for every methods whom return type is a `std::unique_ptr`
-- Use a `std::unique_ptr` as a return type for every methods which has the `SIP_FACTORY` and remove the use of `SIP_FACTORY` for these methods
-- When no methods are using the `SIP_FACTORY` anymore, remove completely the `SIP_FACTORY` definition
+This QEP proposes to use a `std::unique_ptr` as a return type for every methods which has the `SIP_FACTORY` and remove the use of `SIP_FACTORY` for these methods when:
+  - The returned object does not inherit from QObject
+  - The returned object inherits from QObject but NEVER set its parent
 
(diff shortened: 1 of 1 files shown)


rouault:
+1

(Tangential remark to the scope of this PR: the other day I (somewhat naively) tried to change QgsCurvePolygon::setExteriorRing() to *accept* a std::unique_ptr<>. This went well for the C++ part but I got stuck in SIP land where SIP was apparently generating C++ code trying to copy-construct those std::unique_ptr... Apparently related to the fact that this method is virtual. Known issue,d perhaps a SIP one ?)

troopa81:
> (Tangential remark to the scope of this PR: the other day I (somewhat naively) tried to change QgsCurvePolygon::setExteriorRing() to accept a std::unique_ptr<>. This went well for the C++ part but I got stuck in SIP land where SIP was apparently generating C++ code trying to copy-construct those std::unique_ptr... Apparently related to the fact that this method is virtual. Known issue,d perhaps a SIP one ?)

Yes, that's a known issue [fixed](https://github.com/Python-SIP/sip/issues/60) in SIP now . We are almost ready to make the transition (see https://github.com/qgis/QGIS/pull/63814 ) but that require the [Movable](https://python-sip.readthedocs.io/en/stable/annotations.html#mapped-type-annotation-Movable) annotation, only present in SIP 6.11 

But SIP 6.11 hasn't made it yet to all stable Linux distribution (6.10 on Debian trixie for instance). So, we'll have to wait a bit.

nyalldawson:
Do you know how this works for virtual methods, where a C++ method returning a unique_ptr is then overridden in a python subclass?  Does sip correctly handle the ownership and translation of the return value to a C++ uniqur_ptr correctly?

troopa81:
> Do you know how this works for virtual methods, where a C++ method returning a unique_ptr is then overridden in a python subclass? Does sip correctly handle the ownership and translation of the return value to a C++ uniqur_ptr correctly?

You're right... SIP seems to fail to convert the unique_ptr from C++ to Python despite the conversions.sip code. I have to investigate to figure out why.

nyalldawson:
> You're right... SIP seems to fail to convert the unique_ptr from C++ to Python despite the conversions.sip code. I have to investigate to figure out why.

I think it's probably the opposite -- it can't handle the conversion of a new python object returned by the python function to an owned c++ unique_ptr class. Possibly this may magically fix itself when we move to requiring sip 6.11, but I'm doubtful.

Pragmatically I'd adjust the proposal to explicitly exclude any virtual methods which are exposed to Python. Even with that exclusion there's still going to be many methods which will benefit from this cleanup.

nyalldawson:
My other question is how this relates to the QGIS security project crowd funding. If a QGIS funded grant is given for this work, will that count towards the raised funds for that crowdfunding?

m-kuhn:
Good initiative.
Is there anything planned for handling of `QObject` based types (where there's a potential collision of ownership of Qt `parent` and `std::unique_ptr`)?

nyalldawson:
@m-kuhn I don't think we should be annotating those with SIP_FACTORY

m-kuhn:
@nyalldawson I did a quick research upstream (pyqt6) and it seems `QObject` producing methods are annotated with a `/Factory/` directive (at least in some cases):

https://github.com/deepin-community/pyqt6/blob/b140b76543cc2fa882b8f94f6f3dde0da7c20ebd/sip/QtDesigner/customwidget.sip#L38
https://github.com/deepin-community/pyqt6/blob/b140b76543cc2fa882b8f94f6f3dde0da7c20ebd/sip/QtQml/qqmlincubator.sip#L56

nyalldawson:
@m-kuhn 

good point!

So the rule would be something like:

## c++ only methods, not exposed to python

- returning non-qobject: always return unique_ptr
- creturning QObject: return by raw pointer, UNLESS the parent object can NEVER be set for the returned 

## methods exposed to python

- non-virtual methods, returning non-qobject: always return unique_ptr
- virtual methods, return non-qobject: return raw pointer, annotate with SIP_FACTORY or SIP_TRANSFERBACK
- methods returning qobject: always return by raw pointer, annotate with SIP_FACTORY or SIP_TRANSFERBACK

?

troopa81:
> I think it's probably the opposite -- it can't handle the conversion of a new python object returned by the python function to an owned c++ unique_ptr class. Possibly this may magically fix itself when we move to requiring sip 6.11, but I'm doubtful.

Yes, this is from python to cpp. 

> Pragmatically I'd adjust the proposal to explicitly exclude any virtual methods which are exposed to Python. Even with that exclusion there's still going to be many methods which will benefit from this cleanup.

I'll take time to investigate a bit further, see if there is a proper solution (or at least open an issue in SIP so it could be fixed later). But in the meantime, I'm gonna update the QEP to reflect this issue.

> @nyalldawson I did a quick research upstream (pyqt6) and it seems QObject producing methods are annotated with a /Factory/ directive (at least in some cases):

Good catch!

> So the rule would be something like:

So the rule **for method transferring ownership to the caller** would be something like:

> non-virtual methods, returning non-qobject: always return unique_ptr

non-virtual methods, returning non-qobject or **QObject object that are never parented**: always return unique_ptr

> virtual methods, return non-qobject: return raw pointer, annotate with SIP_FACTORY or SIP_TRANSFERBACK

virtual methods ~~, return non-qobject~~: return raw pointer, annotate with SIP_FACTORY or SIP_TRANSFERBACK

I propose to update our coding rules with those rules during the execution of the QEP.

troopa81:
> My other question is how this relates to the QGIS security project crowd funding. If a QGIS funded grant is given for this work, will that count towards the raised funds for that crowdfunding?

Both are trying to achieve the same goal here, improving memory management, that's why I mentioned Security project for QGIS in the QEP. But I think I would prefer to have it separated for the sake of clarity. It's a QGIS.org funded separated program.

nyalldawson:
Isn't one item in the security project explicitly "Improve Memory Safety". How is this different to that?

troopa81:
> Isn't one item in the security project explicitly "Improve Memory Safety". How is this different to that?

It's not. Like I said this try to achieve the same thing.

I have no strong opinion here, I just wanted to not blur the lines between the 2 initiatives. I want the people be completely aware that this is QGIS.org money/initiative/program (if it gets selected obviously). But I'm open to consider it in a different way.

m-kuhn:
Is it feasible to use [QObjectUniquePtr](https://github.com/qgis/QGIS/blob/2886d6ebc082e67d4b0c6eb7ec7648a0646dab95/src/core/qobjectuniqueptr.h#L82) here?

troopa81:
> Is it feasible to use [QObjectUniquePtr](https://github.com/qgis/QGIS/blob/2886d6ebc082e67d4b0c6eb7ec7648a0646dab95/src/core/qobjectuniqueptr.h#L82) here?

I'm not sure. QObjectUniquePtr is convenient when you have a QObject parented object **and** you still want to delete it at the end of your scope. For instance: QgsRubberBand parented to QgsMapCanvas but you delete it when the map tool gets destroyed. But I don't think that's the case for the method like this [one](https://github.com/qgis/QGIS/blob/d0e537f550fd433e67315ded6706f90cbeb1078d/src/core/browser/qgsdataitemprovider.h#L68)  or the one you [mentioned](https://github.com/deepin-community/pyqt6/blob/b140b76543cc2fa882b8f94f6f3dde0da7c20ebd/sip/QtDesigner/customwidget.sip#L38):
- if parentItem is null, the caller is the ownership and you'll have to delete it. So QObjectUniquePtr is ok.
- if parentItem is not null, your object is already parented and you may or may not want to delete it at the end of the scope. 90% the time, I'd say you don't want to delete it, and so you'll have to call release() all the time although you just wanted to have a pointer on an already parented object.

nyalldawson:
Another consideration -- we should always make sure we use Qt's parent/child ownership for Qobjects with QObject members, even if a unique_ptr would be appropriate to handle the ownership instead of Qt's model. This is because we often move objects between threads using QObject::moveToThread, and that only correctly moves member objects which are part of the Qt parent/child model. If we used unique_ptr for those, we'd leave the member on the old thread, resulting in some very hard to diagnose threading issues....

troopa81:
> Another consideration -- we should always make sure we use Qt's parent/child ownership for Qobjects with QObject members, even if a unique_ptr would be appropriate to handle the ownership instead of Qt's model. This is because we often move objects between threads using QObject::moveToThread, and that only correctly moves member objects which are part of the Qt parent/child model. If we used unique_ptr for those, we'd leave the member on the old thread, resulting in some very hard to diagnose threading issues....

Out-of-curiosity, do you a specific QGIS issue in mind or some readings about this ?

Shall we define a rule **Every QObject class member owned by this class must not be a unique_ptr and parented to the class** ? (I'd like maybe write down what have been discussed here in the coding guidelines)

anitagraser:

qep-417-remove-sip-factory.md
```diff
@@ -0,0 +1,70 @@
...
+On the other hand, QGIS API sometime uses, rightfully, a `std::unique_ptr` as a return type instead of a raw pointer.
+
+Both `std::unique_ptr` and `SIP_FACTORY` describes the same paradigm, making the second redundant. Moreover, QGIS application could benefit from wider use of `std::unique_ptr` to improve memory management.
+
+
+## Proposed Solution
+
+This QEP proposes to use a `std::unique_ptr` as a return type for every methods which has the `SIP_FACTORY` and remove the use of `SIP_FACTORY` for these methods when:
+  - The returned object does not inherit from QObject
+  - The returned object inherits from QObject but NEVER set its parent
```
nyalldawson, line 30: Can you tweak this to reflect the "always set qobject parent for owned child objects" rule as discussed?

troopa81:
> I think it's probably the opposite -- it can't handle the conversion of a new python object returned by the python function to an owned c++ unique_ptr class. Possibly this may magically fix itself when we move to requiring sip 6.11, but I'm doubtful.

It was fixed recently, we would have to wait 6.15.4 at least https://github.com/Python-SIP/sip/issues/110#issuecomment-4253744451

qep-417-remove-sip-factory.md
troopa81: I'd like to clarify. 

My sentence was suppose to cover this [case](https://github.com/qgis/QGIS/blob/46cb946405b7e6435fc475aab6e4d864ad5c3e57/src/core/browser/qgsdataitemprovider.h#L68), it takes a parent, so we have to use a SIP_FACTORY and a raw pointer.

But if a method returns a QObject pointer without taking a parent (and is not a virtual/override method), then we should use a unique_ptr, like [here](https://github.com/qgis/QGIS/blob/46cb946405b7e6435fc475aab6e4d864ad5c3e57/src/gui/symbology/qgssymbollayerwidget.h#L206) for instance.

Do we agree on that ? Shall we use a SIP_FACTORY in second case too? If so, why?

nyalldawson: > But if a method returns a QObject pointer without taking a parent (and is not a virtual/override method), then we should use a unique_ptr, like [here](https://github.com/qgis/QGIS/blob/46cb946405b7e6435fc475aab6e4d864ad5c3e57/src/gui/symbology/qgssymbollayerwidget.h#L206) for instance.
Do we agree on that ? Shall we use a SIP_FACTORY in second case too? If so, why

Good example! I think for that situation returning a unique_ptr is correct (or maybe... QObjectUniquePtr, but that probably just causes more sip issues then it's worth :laughing: )

troopa81: So, my sentence `The returned object inherits from QObject but NEVER set its parent` explicit correctly what we want here no? What am I missing ?

PR #375, opened 2026-04-09
QEP 418: Better versioning for ".model3" file format

ValentinBuira:
This QEP aims to improve the versioning of  `.model3` files used to store processing models, and brings it to the same level of standards as we know for the `.qgs` format 

See the file `qep-418-better-versioning-model3-files.md`

This QEP will be submitted as a part of the [QGIS grants program](https://blog.qgis.org/2026/03/16/qgis-grants-11-call-for-grant-proposals-2026/)


qep-418-better-versioning-model3-files.md
```diff
@@ -0,0 +1,136 @@
...
+## Proposed Solution
+
+We propose to introduce new information about the QGIS version the model was made in directly inside the “.model3” format. The idea is directly drawn from the existing logic around the .qgs file format. 
+
+
+## A sense of déjà vu ? 
+
+From the perspective of the user, the UI will be very similar of the warning already existing when opening a QGIS project (same terminology, color code, in message bar, on open). And similar to the project. Saving could effectively change the “.model3” file.
+
+**When loading a model created in a new version** :  “This model file was created by a newer version of QGIS (4.XX.X) and could not be completely loaded.”
```
m-kuhn:
```suggestion
**When loading a model created in a new version** :  “This model file was created by a newer version of QGIS (4.XX.X) and might not be fully compatible with this version.”
```

I think we should improve this also for QGIS, this bears the meaning that we were not able to load it completely. This is not correct, it's well possible we were able to completely load it, we are just not sure.

qep-418-better-versioning-model3-files.md
```diff
@@ -0,0 +1,136 @@
...
+   mDesignerParameterValues = map.value( u"designerParameterValues"_s ).toMap();
+```
+
+
+Finally QgsProcessingModelAlgorithm::toVariant will be updated to always save into the “.model3” the current version of QGIS, and the format supported by this version.
+
+
+## Continuous process
+
+From now on, for each release where we make breaking changes to the model in QGIS we would have to add a new versioning condition. This would be ensured during review of processing related PR.
```
m-kuhn: Can you describe the "versioning condition" a bit more.

qep-418-better-versioning-model3-files.md
```diff
@@ -0,0 +1,136 @@
...
+* Open a model with a feature added in later version of QGIS is broken, for example you open a model made in QGIS 3.44 that use [repeat expression](https://qgis.org/project/visual-changelogs/visualchangelog344/#feature-add-repeat-function-and-reverse-variant-for-strings) in QGIS 3.40, same could be said for any new processing algorithms
+
+
+Both case silently broke the model open, without a clear warning.
+
+And in the first case, _we effectively silently break the promise of backward compatibility when it comes to the processing model._ 
+
+This QEP proposes to reestablish the promise of backward compatibility in the model designer. And conveniently introduce a framework for a better versioning in the “.model3” and allow us to make the format evolving as much as we want. 
+
+For example one thing that is missing in the “.model3” format is the dependency to plugins. If a model is using an algorithm from a plugin and the plugin is not installed. The only information we have is the name of the provider of the plugin.
```
m-kuhn: Is this addressed by this QEP?

ValentinBuira: > Is this addressed by this QEP?

No but it's an example of what would be available to us. This QEP would rather focus on finding out what were the breaking changes over the last years (apart from the one from in the example), and fix them

m-kuhn: Can you add a sentence, how this relates to this QEP (would that benefit from this framework or is it otherwise affected)?

qep-418-better-versioning-model3-files.md
```diff
@@ -0,0 +1,136 @@
...
+_quick'n dirty mockup to give a rough idea how that would look_ 
+
+![](./images/qep418/mockup-user-feedback.png)
+
+## A new key information the QGIS version
+
+A new entry will be introduced in the file format corresponding to the version( i.e `Qgis::version()` ) in which the model was last saved. This version information will serve as the pivot point to know how to load the model file. 
+
+
+> To note the “.model3” file format already went to a format change around QGIS 3.26, as you can see with the enum `QgsProcessingModelAlgorithm::InternalVersion`. But only regarding the structure of the documents and not the version of QGIS.
```
troopa81: > But only regarding the structure of the documents and not the version of QGIS.

Not sure to fully understand this, It seems completely related to the QGIS version when I read the enum comments

 >    Version1, //!< Created in < 3.26
  >   Version2, //!< Created in >= 3.26
 
Why not replacing InternalVersion with just the QGIS version?

qep-418-better-versioning-model3-files.md
```diff
@@ -0,0 +1,136 @@
...
+While QGIS is evolving the “.model3” file remains static. This lead to multiple problems:
+
+
+* Deprecated or renamed parameters like  https://github.com/qgis/QGIS/commit/5bde6fec42ccc10b9fd1c4fc822fbe1a4ae05a30#diff-1a68bfeb89b1356e8ae1095b1d3576c138b2809adde3dbd2a123819c2b3a33b6R108, break the model. the idea of this QEP actually emerged after I wanted to change the type of an output from string to CRS.
+* Open a model with a feature added in later version of QGIS is broken, for example you open a model made in QGIS 3.44 that use [repeat expression](https://qgis.org/project/visual-changelogs/visualchangelog344/#feature-add-repeat-function-and-reverse-variant-for-strings) in QGIS 3.40, same could be said for any new processing algorithms
+
+
+Both case silently broke the model open, without a clear warning.
+
+And in the first case, _we effectively silently break the promise of backward compatibility when it comes to the processing model._
```
troopa81, line 24: Why is it breaking backward compatibility ? With the modification, you would still be able to load older models with new version of QGIS, and it would rely on older `OPTIONS` not the new `CREATION_OPTIONS` one.

This is because you'd like to get rid of the old `OPTIONS` one and keep only the new one ? But doing so would also break calling this algorithm using qgis_process.

qep-418-better-versioning-model3-files.md
```diff
@@ -0,0 +1,136 @@
...
+**Version** QGIS 4.4
+
+## Summary
+
+QGIS uses a “.model3” file format for storing models created using the model designer. The “.model3” despite the extension name is just an xml document under disguise.
+
+While QGIS is evolving the “.model3” file remains static. This lead to multiple problems:
+
+
+* Deprecated or renamed parameters like  https://github.com/qgis/QGIS/commit/5bde6fec42ccc10b9fd1c4fc822fbe1a4ae05a30#diff-1a68bfeb89b1356e8ae1095b1d3576c138b2809adde3dbd2a123819c2b3a33b6R108, break the model. the idea of this QEP actually emerged after I wanted to change the type of an output from string to CRS.
```
wonder-sk, line 18: How does the linked commit break model? From a quick look it seems the switch to a new parameter was done with backward compatibility in mind...?

qep-418-better-versioning-model3-files.md
```diff
@@ -0,0 +1,136 @@
...
++
++    
++    push the info about the model version to the user 
++
++    if (fileVersion < version 3.44): 
++      //  Transform from old version to new version
++          fileVersion = fileVersion 4.0
++    
++    if (fileVersion <  version 4.2):
++      //  Transform from version 4 to version 4.2
```
wonder-sk: This is a déjà vu too 🙂 We have a mechanism for migration of QgsProject files - see QgsProjectFileTransform. Introduced around 18 years ago, there are only roughly three transforms implemented - i.e. it has ended up pretty much unused - and it has been more convenient to handle the project file backwards compatibility in the bit of the code where things are actually parsed. My gut feeling is that we would end up with the same thing here as well.

And when dealing with stuff like parameter renames, it feels like those should be handled within individual algorithm implementations anyway, rather than handled globally...

qep-418-better-versioning-model3-files.md
alexbruy: I also would be interested in getting more info on how this breaks models. This approach is used in many algorithms and it was designed to keep backward compatibility.

ValentinBuira: We don't expose deprecated/hidden parameters in the model designer, 

So if we take a simple example model , I made in 3.40 using the former "OPTIONS" parameter (it's just an example btw this model means nothing) 

In 3.40

Screenshot 2026-04-14 125727


The same model in 3.44.8

Screenshot 2026-04-14 125813


The link to the "OPTIONS" parameter have disappear, and the actual outcome of the model is left to the details of implementation. I am not 100% sure of what happen here on execution. Is the former parameter used but not editable (if it was not connect for example and just used as a simple string )? is it ignored ? What if I save my model is the hidden parameter lost ?


In short it break the model but not the python API you can still do this in python:

```
processing.run(
    "gdal:cliprasterbyextent",
    {
        "INPUT": "C:/Users/Valentin/QField/cloud/oulipo_rennes_qfield_1/20250719_170012_modified.tif",
        "PROJWIN": "-187063.789800000,-184837.115300000,6123700.529600000,6125144.931100000 [EPSG:3857]",
        "OVERCRS": False,
        "NODATA": None,
        "OPTIONS": "COMPRESS=PACKBITS", # DEPRECATED but still work in PyQGIS running 3.44.8
        "DATA_TYPE": 0,
        "EXTRA": "",
        "OUTPUT": "TEMPORARY_OUTPUT",
    },
)
```

qep-418-better-versioning-model3-files.md
ValentinBuira: See my response in the other comment : https://github.com/qgis/QGIS-Enhancement-Proposals/pull/375#discussion_r3078969124

qep-418-better-versioning-model3-files.md
wonder-sk: I see - so the problem is that the model looses (?) the existing connection if a parameter is hidden in the new version. In that case, my suggestion would be to have some new API within processing algorithm class that could tell the processing framework how to migrate existing connections, with several options, e.g.:
- connect to parameter X instead
- do not connect, emit a warning or an error

ValentinBuira: I did more tests with manually setting a parameter of the buffer algorithm to hidden ( the tests was conducted in a private branch by just switching a parameter to hidden, no more logic was added to the processAlgorithm).

Even if not linked hidden parameters are an issue since they are executed anyway, It would be more accurate to describe them as "ghost parameters"

> new API within processing algorithm class that could tell the processing framework 
> how to migrate existing connections, with several options, 

I think it would make sense to have an API at individual level like you said, this way algorithms from plugins could also benefits from it. 

In fact _I think both this QEP and a new API would complement each others very well._ because this type of API would still break the model for older version of QGIS, we need to store info in the model we don't at the moment etc..

For the new API this raise new questions for after this QEP, should we have a internal version number per algorithms ? per provider ?

qep-418-better-versioning-model3-files.md
ValentinBuira: > And when dealing with stuff like parameter renames, it feels like those should be handled within individual algorithm implementations anyway, rather than handled globally...

For other readers,  see in the other comment https://github.com/qgis/QGIS-Enhancement-Proposals/pull/375#discussion_r3100752820

qep-418-better-versioning-model3-files.md
ValentinBuira: It's really close, as far as  I understand it, InternalVersion still maintains compatibility with both version with `switch/case` in the code to handle both version, while in this proposals a model would be transformed to latest version only

But on second thought maybe we could scrap this and rely only on the QGIS version introduced by this QEP ?

anitagraser:

nyalldawson:
@ValentinBuira can you update the proposal text to reflect the outcome of the discussions before we start putting in votes on this one?

qep-418-better-versioning-model3-files.md
ValentinBuira: Added a paragraph about change that would need to be versioned by this logic.

Since the discussion raised that the versioning of algorithms should rather happen at the level of algorithms themselves (see the discussion below). The number of format wide changes is now much lower

ValentinBuira:
> @ValentinBuira can you update the proposal text to reflect the outcome of the discussions before we start putting in votes on this one?

I updated the proposal. The main change is the need for an additional API within the processing algorithm class to tell QGIS how the versioning of the algorithm himself should be handled when loading a model in QGIS

It is now in a state ready for voting

PR #376, opened 2026-04-10
QEP 419: Improved Wayland compatibility

nyalldawson:
Over the last couple of years, most Linux distributions and desktop environments have been transitioning from the mature X11 display server to the Wayland standard. The transition has reached a stage where many environments are now completely dropping support for the X11 server.

This presents a challenge to Linux QGIS users, as the current QGIS versions have limitations and bugs when run in a Wayland environment.

This QEP aims to address these limitations in the "best possible" way, given the (many!) constraints of Wayland.

(Unfortunately, the Wayland maintainers are strongly opinionated have very strict ideas about what a modern application should be allowed to do, and their vision does not fit well with the needs of QGIS. While it will not be possible to completely support the full capabilities and user workflows that are possible in X11, there are opportunities to fix some of the current shortcomings. :upside_down_face: )

A rendered version of the proposal is available at https://github.com/nyalldawson/QGIS-Enhancement-Proposals/blob/2141576af4de883a564f9673c7f314d958d25a48/qep-419-improved-wayland-compatibility.md

### DIFF
--- a/qep-419-improved-wayland-compatibility.md
+++ b/qep-419-improved-wayland-compatibility.md
@@ -11,7 +11,7 @@
 # Summary
 
 Over the last couple of years, most Linux distributions and desktop environments have been transitioning
-from the legacy X11 display server to the Wayland standard. The transition has reached a stage where
+from the mature X11 display server to the Wayland standard. The transition has reached a stage where
 many environments are now completely dropping support for the X11 server.
 
 This presents a challenge to Linux QGIS users, as the current QGIS versions have limitations and bugs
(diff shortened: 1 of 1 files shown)


lnicola:
As a light QGIS user, Wayland support has been rock-solid in pre-4.0 for a good number of releases (despite the scary warnings), but I don't use the "problematic" features. More so, I _don't want_ QGIS to be able to take screenshots or sample screen pixels without my knowledge. I _don't want_ QGIS plugins to be able to take screenshots either.

Coming back to the RFC, my suggestion is to nudge the Linux developers into actually running Wayland. I encountered some ugly Qt warnings in 4.0 -- a Qt update seems to have fixed most, but not all of them.

nyalldawson:
@lnicola 
> Coming back to the RFC, my suggestion is to nudge the Linux developers into actually running Wayland. I encountered some ugly Qt warnings in 4.0 -- a Qt update seems to have fixed most, but not all of them.

It's already the case. All linux developers using up to date gnome have already had wayland forced on them. 

For myself, I always test QGIS under wayland on the newest Qt /KDE releases, and there's STILL severe issues remaining. My daily driver is X11 as a result.

anitagraser:

uclaros:
+1

alexbruy:
+1

PR #378, opened 2026-04-13
QEP 421: Add help strings for processing algorithm parameters

uclaros:
After having resorted to checking the source code to find out what a processing algorithm's parameter does, I am convinced we should widely adopt using `QgsProcessingParameterDefinition::setHelp()`.

For QGIS grant 2026, I propose to add all missing help messages to processing algorithm parameters and try to force using them in new algorithms.

Check the details in https://github.com/uclaros/QGIS-Enhancement-Proposals/blob/processing-param-sethelp/qep-421-processing-parameter-set-help.md

### DIFF
--- a/qep-421-processing-parameter-set-help.md
+++ b/qep-421-processing-parameter-set-help.md
@@ -17,27 +17,26 @@ Tooltips are arguably the most accessible way to communicate to users some extra
 In QGIS 3.16 custom help messages for parameters displayed on tooltips was introduced. These get defined using `QgsProcessingParameterDefinition::setHelp( const QString &help )` for each instantiated parameter.
 Many new algorithms use the `setHelp()` method for their parameters, offering a nice UX to users, however its usage was not globally adopted, resulting in lots of parameters with no description.
 
+On the other hand, the QGIS documentation pages have a rich collection of processing algorithm parameter descriptions.
 
 
 ## Proposed Solution
-We propose to go through all processing algorithms, identify the parameters' role and include a translatable string as a help message for all parameters that do not have one.
+We propose to go through all processing algorithms and transfer the parameters' descriptions from the QGIS documentation pages into `setHelp()` calls to make them available in tooltips.
+Some descriptions may require modification to fit the context of the processing algorithm tooltips, eg. for combo boxes the description of their funtion is required but the available values are not.
 
-We also suggest to add the help string as an optional parameter in all `QgsProcessingParameterDefinition` subclasses constructors. This will allow for creating a CI test to force using help strings for all parameters, without breaking the API for the Processing Parameter subclasses constructors.
+We also suggest to add a CI test to force using help strings for all parameters, excluding input and output ones. This will ensure that new algorithms have proper descriptions for their parameters which the documenters will be able to reuse for the documentation repo.
 
 ## Deliverables
 
-- Modified constructors for `QgsProcessingParameterDefinition` subclasses
-- Updated `initAlgorithm()` methods in processing algorithms to include the parameters' new help string
+- Updated `initAlgorithm()` methods in processing algorithms to include the parameters' new help strings
 - CI test that will fail when processing parameters without a help string are created.
 
 ### Affected Files
 
-- `src/core/processing/qgsprocessingparameters.[cpp|h]`
-- potentially constructors of other `QgsProcessingParameterDefinition` subclasses in seperate files
-`initAlgorithm` methods in:
 - `src/analysis/processing/*.cpp
 - `python/plugins/processing/algs/gdal/*.py`
 - `python/plugins/processing/algs/qgis/*.py`
+- `python/plugins/processing/tests/AlgorithmsTestBase.py`
 
 ## Risks
 
@@ -49,7 +48,8 @@ We also suggest to add the help string as an optional parameter in all `QgsProce
 N/A
 
 ## Further Considerations/Improvements
-In QGIS 5 we can consider changing the API to require both `description` and `help` for all `QgsProcessingParameterDefinition` subclasses.
+
+For parameters that lack a proper description, a new translatable string will be added and pushed to the documentation repo, too.
 
 
 ## Backwards Compatibility
--- a/qep-421-processing-parameter-set-help.md
+++ b/qep-421-processing-parameter-set-help.md
@@ -1,0 +1,57 @@
+# QGIS Enhancement: Force usage of help strings for processing algorithm parameters
+
+**Date** 2026/04/13
+
+**Author** Stefanos Natsis (@uclaros)
+
+**Contact** uclaros at gmail dot com
+
+**Version** QGIS 4.4
+
+# Summary
+
+The labels of processing algorithm parameters are not always self explenatory. In the past, I have resorted to reading the source code in order to understand what a parameter does.
+While the algorithms' help messages may include descriptions for the parameters, it is often inadequate or may even be overlooked by the users.
+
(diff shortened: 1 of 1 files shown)


qep-421-processing-parameter-set-help.md
```diff
@@ -0,0 +1,57 @@
...
+
+- A bulk of new strings for translation will be introduced, adding extra work to the translators.
+- We may end up with redundant help messages for self explanatory parameters
+
+## Performance Implications
+
+N/A
+
+## Further Considerations/Improvements
+In QGIS 5 we can consider changing the API to require both `description` and `help` for all `QgsProcessingParameterDefinition` subclasses.
```
rouault: would a CI task that would enumerate all parameters of QGIS built-in algorithms and do that check be easily doable to ensure we don't regress after this sanitization work has been done, without waiting for a unplanned QGIS 5 ?

uclaros: This is the short term plan indeed:

> We also suggest to add the help string as an optional parameter in all `QgsProcessingParameterDefinition` subclasses constructors. This will allow for creating a CI test to force using help strings for all parameters, without breaking the API for the Processing Parameter subclasses constructors.

I think that having the help string in the constructor will facilitate the CI task, but I'm open to alternative suggestions.

qep-421-processing-parameter-set-help.md
```diff
@@ -0,0 +1,57 @@
...
+Tooltips are arguably the most accessible way to communicate to users some extra information about those parameters.
+In QGIS 3.16 custom help messages for parameters displayed on tooltips was introduced. These get defined using `QgsProcessingParameterDefinition::setHelp( const QString &help )` for each instantiated parameter.
+Many new algorithms use the `setHelp()` method for their parameters, offering a nice UX to users, however its usage was not globally adopted, resulting in lots of parameters with no description.
+
+
+
+## Proposed Solution
+We propose to go through all processing algorithms, identify the parameters' role and include a translatable string as a help message for all parameters that do not have one.
+
+We also suggest to add the help string as an optional parameter in all `QgsProcessingParameterDefinition` subclasses constructors. This will allow for creating a CI test to force using help strings for all parameters, without breaking the API for the Processing Parameter subclasses constructors.
```
nyalldawson: What does this mean? Isn't it already an optional parameter, as the creator needs to manually call param->setHelp("xxx") ?

nyalldawson:
For full disclosure -- would the help strings be generated by hand, or by LLM? I'm not -1 to LLM use here (I've personally found gemini is quite good for populating them), but they'd still need to be reviewed one-by-one if generated in that way.

uclaros:
> For full disclosure -- would the help strings be generated by hand, or by LLM? I'm not -1 to LLM use here (I've personally found gemini is quite good for populating them), but they'd still need to be reviewed one-by-one if generated in that way.

The plan is to walk through all parameters and generate the strings by hand, I'll update the proposal to reflect that. An LLM might still be used to correct bad English and potentially for generating `INPUT` and `OUTPUT` parameter strings, I am not settled on whether we should skip those and exclude them from the CI run though... Any thoughts on that?

nyalldawson:
> I am not settled on whether we should skip those and exclude them from the CI run though... Any thoughts on that?

Good question. I'd skip them -- I think otherwise it'd just lean towards low quality unhelpful strings ("is the input layer"/"is the output layer")

qep-421-processing-parameter-set-help.md
uclaros: The idea is to include them as parameters _in the constructors_, as that would make the CI checking easier.
Currently, the help message exists in parent class constructor:
```cpp
QgsProcessingParameterDefinition::QgsProcessingParameterDefinition( const QString &name, const QString &description, const QVariant &defaultValue, bool optional, const QString &help )
  : mName( name )
  , mDescription( description )
  , mHelp( help )
  , mDefault( defaultValue )
  , mFlags( optional ? Qgis::ProcessingParameterFlag::Optional : Qgis::ProcessingParameterFlag() )
{}
```

however, it is not really used in any of the subclasses constructors:
```cpp
QgsProcessingParameterBoolean::QgsProcessingParameterBoolean( const QString &name, const QString &description, const QVariant &defaultValue, bool optional )
  : QgsProcessingParameterDefinition( name, description, defaultValue, optional )
{}
```

nyalldawson: Couldn't a ci check just iterate through all algorithms, iterate through the parameters , and check that help string is not empty?

uclaros: D'oh! As a proper cpp test! That just makes sense!
My mind was focused on building a parser script instead.

Only drawback is we won't be able to have some `// skip-keyword-check` equivalent. We'll be just skipping 'INPUT', 'OUTPUT' etc.

ValentinBuira:
Hi @uclaros 

The built-in documentation definitely deserve some love. And worst thing is sometimes the documentation is already existing but in the online documentation ( https://docs.qgis.org/3.44/en/docs/user_manual/processing_algs/qgis/index.html )

For example take the buffer algorithms and and if you compare, the tooltip, side docs, and online doc, see the text for the "DISTANCE" parameter or either the "END_CAP_STYLE" with visual is very instructive

built-in 

Screenshot 2026-04-14 121351

Online 

Screenshot 2026-04-14 121649


<br><br>

Have you consider pulling the strings from the docs to make the built-in and online doc converge ? 

And that would be even better if we could somehow convert the png image in the doc to svg to display then in the sidebar doc too ? But  I am thinking out loud here

baswein:
> And that would be even better if we could somehow convert the png image in the doc to svg to display then in the sidebar doc too ?

I think I made those images and at the time thought that it would be nice to have them in the algorithm help text but didn't see that happening anywhere else. Also I think they were originally svg's so it might make sense to have them and others as svg's in the documentation if that would facilitate them being included in the algorithm help text.

qep-421-processing-parameter-set-help.md
nyalldawson: I'd add the tests to https://github.com/qgis/QGIS/blob/master/python/plugins/processing/tests/AlgorithmsTestBase.py#L554 -- those are similar tests.

And then just have a hardcoded list of allowed exceptions in the test itself.

uclaros:
Thanks @ValentinBuira for bringing up the documentation site, I keep forgetting that the `help` button takes you to the proper online documentation page for the algorithm and not some generic help viewer made for windows 3.11!

Looking at the documentation pages, the vast majority of parameters are well documented, so pulling those strings to the algorithms' sources really makes sense and reduces the required effort.

I'm not sure if building some syncing mechanism between code help and docs would be worth it, as these descriptions are not expected to be modified much in the future. I will give this some more thought and update the proposal, possibly also the quoted time.

nyalldawson:
Fwiw, I've always thought it made sense for the help strings in qgis to be the definitive versions, with the documentation syncing from those.

anitagraser:

nyalldawson:
@uclaros can you update the proposal text to reflect the discussion results?

uclaros:
### Pitch

This proposal aims to improve the usability of existing processing algorithms by populating all their parameters' missing tooltips with their descriptions from the online documentation. New description strings will be added for parameters not covered by the documentation. A test will also be added to guarantee that no new algorithms/parameters will be introduced in the future without their help strings populated.

nyalldawson:
+1

alexbruy:
+1

PR #379, opened 2026-04-13
QEP 422: Async Refactoring

dvdkon:
In this QEP, I propose refactoring the usage of async primitives in QGIS to make use of new features from Qt6, namely QPromise and QFuture chaining methods. This should improve code readability and fix some edge-case bugs.

This QEP is submitted as part of a QGIS Grant Application for 2026.

### DIFF
--- a/qep-422-async-refactoring.md
+++ b/qep-422-async-refactoring.md
@@ -1,0 +1,71 @@
+# QGIS Enhancement: Async Refactoring
+
+**Date** 2026/04/13
+
+**Author** David Koňařík (@dvdkon)
+
+**Contact** dvdkon at konarici dot cz
+
+**Version** QGIS 4.4
+
+# Summary
+
+QGIS does a lot of operations asynchronously, to not block the main thread and
+to improve throughput. Currently, those operations are handled by low-level
+primitives, like threads, signals and at best a `QFutureWatcher`. This approach
+is complex and bug-prone, as can be seen in e.g. [this 3D tile loading
+issue](https://github.com/qgis/QGIS/issues/63340).
+
+In Qt6, better and more modern primitives for writing asynchronous code [were
+added](https://www.qt.io/blog/asynchronous-apis-in-qt-6), like `QPromise` and
+new methods on `QFuture`. Having upgraded to Qt6, we can now use these to
+refactor our async code using paradigms proven by use in other languages and
+frameworks.
+
+## Proposed Solution
+
+The proposed refactoring will replace uses of raw signals and `QFutureWatcher`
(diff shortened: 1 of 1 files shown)


nyalldawson:
Sounds good to me!

Guts:
Thank you for this exciting proposal, which will help modernize the QGIS codebase.

Is any of this exposed in the Python API in some way? Currently, plugin developers mainly use QgsTasks and everything that goes with them ([subtasks](https://qgis.org/pyqgis/master/core/QgsTask.html#module-QgsTask), [manager](https://qgis.org/pyqgis/master/core/QgsTaskManager.html#module-QgsTaskManager), etc.).
Would this serve as a replacement or a supplement?

dvdkon:
> Is any of this exposed in the Python API in some way? Currently, plugin developers mainly use QgsTasks and everything that goes with them ([subtasks](https://qgis.org/pyqgis/master/core/QgsTask.html#module-QgsTask), [manager](https://qgis.org/pyqgis/master/core/QgsTaskManager.html#module-QgsTaskManager), etc.). Would this serve as a replacement or a supplement?

`QgsTask` is a more user-facing concept that holds more data, so it couldn't be replaced by `QFuture` wholesale. Also SIP/PyQt sadly doesn't support `QFuture`, so until support is added, we have to keep our public APIs signals-based. The best we could do is allow constructing tasks from futures, if that is useful anywhere.

troopa81:
There are so many way of doing thread synchronization in C++/Qt (classic c++ one, QTread/QMutex, QWaitCondition , QRunnable, QtConcurrent, QFuture, using local event loop ...) used in the whole QGIS code base that I'm a bit concerned about adding a new one (especially a Qt one). 

But I have never experienced it so I would not oppose if other contributors agree that it would simplify code and make it less error prone. 

But shall we not establish some coding guidelines about what to use depending on the scenarios?

anitagraser:

dvdkon:
> There are so many way of doing thread synchronization in C++/Qt (classic c++ one, QTread/QMutex, QWaitCondition , QRunnable, QtConcurrent, QFuture, using local event loop ...) used in the whole QGIS code base that I'm a bit concerned about adding a new one (especially a Qt one).

Since we're already using Qt heavily in QGIS, I think sticking to the Qt options make sense. Within those, the higher-level ones build on lower-level concepts, so they're not really duplicates. I think choosing the highest-level abstraction that still makes sense is a reasonable guideline.

nyalldawson:
+1

alexbruy:
+1

nyalldawson:
Declared as approved (on a technical basis) with votes from @nyalldawson , @alexbruy

PR #380, opened 2026-04-13
QEP 423: Get rid of QgsProject::instance() singleton in qgis_core

wonder-sk:
This QEP propose to continue the efforts started long time ago to get rid of the QgsProject singleton usage in the qgis_core library to prevent API users from unpleasant surprises...

See the formatted version here: https://github.com/wonder-sk/QGIS-Enhancement-Proposals/blob/qgsproject-instance/qep-423-qgsproject-instace-qgis-core.md

This QEP is submitted as part of a QGIS Grant Application for 2026.

### DIFF
--- a/qep-423-qgsproject-instace-qgis-core.md
+++ b/qep-423-qgsproject-instace-qgis-core.md
@@ -1,0 +1,58 @@
+# QGIS enhancement: Get rid of `QgsProject::instance()` singleton in qgis_core
+
+**Date** 2026-04-13
+
+**Author** Martin Dobias (@wonder-sk)
+
+**Contact** wonder dot sk at gmail dot com
+
+**Version** QGIS 4.4
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


nyalldawson:
Good initiative!

I wonder if it'd be better to change the project from "get rid of" to "work towards getting rid of", and then putting a hour cap in the proposal? There's definitely going to be some nasty surprises here which will burn a lot of hours... so to protect yourself it's probably better to have a definite "project complete" milestone for when your budgeted hours are exhausted.

anitagraser:

NathanW2:
Excellent!  Good old singletons, so handy at the time but such a footgun for the future.

wonder-sk:
@nyalldawson 

> I wonder if it'd be better to change the project from "get rid of" to "work towards getting rid of", and then putting a hour cap in the proposal? There's definitely going to be some nasty surprises here which will burn a lot of hours... so to protect yourself it's probably better to have a definite "project complete" milestone for when your budgeted hours are exhausted.

Sorry I have forgotten to reply - the deliverables section actually has a note about that:

_The target is to get to zero occurences within qgis_core library, but due to the lack of time a couple of hard-to-fix cases may be left for future fixes._

nyalldawson:
+1

alexbruy:
+1

PR #382, opened 2026-05-07
Add QEP 425: Stable API policy

### LINKED ISSUE #321: Document stable api guarantee requirements
This would be another good candidate for a policy style QEP

nyalldawson:
This QEP documents policies for API stability for QGIS.

It is initially intended as a formal description of CURRENT (undocumented) POLICIES ONLY, which can be later modified and revised (in an atomic manner) if we want to change any of these policies. As such, please refrain from using this initial request as a place to discuss changes to current practice. Rather, let's keep discussion focused on whether this is an accurate and complete description of our current policies. 👍

The end goal here is having a formal policy in place describing all our API stability rules, which will help ease new developers into contributing effectively to QGIS and remove some of the unspoken assumptions surrounding QGIS development.

Fixes https://github.com/qgis/QGIS-Enhancement-Proposals/issues/321

### DIFF
--- a/qep-425-stable-api.md
+++ b/qep-425-stable-api.md
@@ -1,0 +1,57 @@
+# QGIS Enhancement 425: Stable API policy
+
+## Summary
+
+This QEP documents the policy for maintaining a stable QGIS API. Developers must follow these policies
+when submitting changes to the QGIS code.
+
+The stable API policy guarantees that (with exceptions) plugins and scripts written for a particular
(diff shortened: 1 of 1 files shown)


rouault (approved):

troopa81 (commented):
+1

Shall we add also the application customization files (QGIS4.ini and QGISCUSTOMIZATION.XML) to the file format list ?

PR #383, opened 2026-05-27
[CodingStyle] Precise member prefix scope

troopa81:
It applies only to private and protected.

Related to this [comment](https://github.com/qgis/QGIS/pull/66197#issuecomment-4550528015)

### DIFF
--- a/qep-314-coding-style.md
+++ b/qep-314-coding-style.md
@@ -17,7 +17,7 @@ Notes:
 
   - 1.1.1. Local variables should have no name prefix (eg do not prefix variable names with ``my``)
   - 1.1.2. [Hungarian notation](https://en.m.wikipedia.org/wiki/Hungarian_notation) is NOT used, except in some exceptional circumstances (eg variable names used with the gdal c API)
-  - 1.1.3. Private and protected member variables should have a ``m`` prefix (eg ``mLineWidth``)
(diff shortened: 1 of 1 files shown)


qep-314-coding-style.md
```diff
@@ -17,7 +17,7 @@ Notes:
 
   - 1.1.1. Local variables should have no name prefix (eg do not prefix variable names with ``my``)
   - 1.1.2. [Hungarian notation](https://en.m.wikipedia.org/wiki/Hungarian_notation) is NOT used, except in some exceptional circumstances (eg variable names used with the gdal c API)
-  - 1.1.3. Member variables should have a ``m`` prefix (eg ``mLineWidth``)
+  - 1.1.3. Private and protected member variables should have a ``m`` prefix (eg ``mLineWidth``)
```
nyalldawson: Good catch, thanks!

Can we go a bit more explicit and state what we expect for public too? eg 
```suggestion
  - 1.1.3. Private and protected member variables should have a ``m`` prefix (eg ``mLineWidth``). Public member variables should use camel case, without any prefix (eg ``lineWidth``).
```

uclaros: I think we should mention that using public member variables is discouraged and this mostly applies to structs.

troopa81: You are probably right, but I consider this outside of the scope of this proposal which was about the naming of member variable only.

PR #386, opened 2026-07-14
QEP 427: Removal of deprecated Processing GUI API

nyalldawson:
This QEP proposes that the deprecated Processing ["WidgetWrapper" API](https://github.com/qgis/QGIS/blob/master/python/plugins/processing/gui/wrappers.py) (from QGIS 2.x) be removed, requiring all plugins and third party code to use the replacement [QgsAbstractProcessingParameterWidgetWrapper](https://qgis.org/pyqgis/master/gui/QgsAbstractProcessingParameterWidgetWrapper.html) API instead (which was introduced in QGIS 3.4).

As per clause 1.4.4 of the Stable API policy, this API should be removed on the grounds that maintaining backward compatibility both introduces an insurmountable maintenance burden and fundamentally blocks critical bug fixes (condition 1.4.4.1).

See https://github.com/nyalldawson/QGIS-Enhancement-Proposals/blob/28d987a2d338ec131630bc2c11eedaf618f244dd/qep-427-remove-deprecated-processing-gui-api.md for a rendered version of the proposal.


qep-427-remove-deprecated-processing-gui-api.md
```diff
@@ -0,0 +1,109 @@
...
+## Proposed Implementation Plan
+
+### 1. Port remaining QGIS algorithm widgets to new API
+
+There remain a handful of parameter widgets using the old API from the inbuilt QGIS
+algorithms (see https://github.com/qgis/QGIS/tree/master/python/plugins/processing/algs/qgis/ui). Specifically,
+these are:
+
+- The Execute SQL widget
+- The interpolation algorithm custom widgets
```
ValentinBuira, line 41: I would like to raise that the interpolation widget doesn't work in the model designer see https://github.com/qgis/QGIS/issues/30029.

Since this would break the API anyways, could it be an opportunity to redesign/fix this widget parameter at the same time ? 

Although I understand this QEP is not exactly about this API and more about the GUI API. but since the expected format in `processing.run` as `INTERPOLATION_DATA `parameter is also tied to the GUI maybe it's worth considering

nyalldawson: > Since this would break the API anyways, could it be an opportunity to redesign/fix this widget parameter at the same time ?

Actually it's not a API break to do this -- the newer c++ widget wrapper api could definitely handle the interpolation widgets nicely. It's just a matter of effort to do this. (Note that this also applies elsewhere, eg the existing mesh interpolation parameter widget wrapper, which already uses the c++ api, does not offer a model-focused widget either)

